PR #3, opened 2016-05-11
Added initial implementation for Containers

romaninsh:
First pull-request is complete. This one implements traits:

 - TrackableTrait
 - ContainerTrait
 - AppScope


src/AppScopeTrait.php
```diff
@@ -3,5 +3,43 @@
 namespace atk4\core;
 
 trait AppScopeTrait {
+
+    public $_appScope = true;
```
DarkSide666: For consistency correct name of this property would be `$_appScopeTrait = true`. Equal with Trait class name started with underscore and first letter lowercased. Other traits do like this already.

src/AppScopeTrait.php
```diff
@@ -3,5 +3,43 @@
...
+     * Object in Agile Toolkit contain $name property which is derrived from
+     * the owher object and keeps extending as you add objects deeper into
+     * run-time tree. Sometimes that may generate long names. Long names are
+     * difficult to read, they increase HTML output size but most importantly
+     * they may be restricted by security extensions such as SUHOSIN.
+     *
+     * Agile Toolkit implements a mechanism which will replace common beginning
+     * of objects with an abbreviation thus keeping object name length under
+     * control. This variable defines the maximum length of the object's $name.
+     * Be mindful that some objects will concatinate theri name with fields,
```
DarkSide666: I think that objects name concatenating with with field names without shortening resulting name is actually issue of ATK and should be solved where it happens. So I guess this part of comment would be nice to remove :)

src/AppScopeTrait.php
```diff
@@ -3,5 +3,43 @@
...
+     * so the maximum letgth of GET argument names can exceed this value by
+     * the length of your field.
+     *
+     * We recommend you to increase SUHOSIN get limits if you encounter any
+     * problems. Set this value to "false" to turn off name shortening.
```
DarkSide666: Link to Suhosin manual would be nice here. Or at least to StackOverflow like this: http://stackoverflow.com/a/9399615/1466341

src/AppScopeTrait.php
```diff
@@ -3,5 +3,43 @@
 namespace atk4\core;
```
DarkSide666: lacks description of this trait

src/TrackableTrait.php
```diff
@@ -9,14 +9,53 @@
...
 
+    /**
+     * Removes object from parent, so that PHP's Garbage Collector can
+     * dispose of it.
+     */
     function destroy(){
-        unset($this->owner->elements[$this->short_name]);
+        if (
+            isset($this->owner) &&
+            $this->owner->_containerTrait
```
DarkSide666: this will throw warning if owner doesn't have _containerTrait property (or contant) at all. Should use isset() here. Isset "_Returns TRUE if var exists and has value other than NULL, FALSE otherwise._"

src/ContainerTrait.php
```diff
@@ -2,8 +2,33 @@
...
+
+    /**
+     * short_name => object hash of children objects. If the child is not
+     * trackable, then object will be set to "true" (to avoid extra reference)
+     *
+     * @var array
+     */
     protected $elements;
 
     private $_element_name_counts = array();
```
DarkSide666: method _unique_element() will fail if $desired === null

src/ContainerTrait.php
```diff
@@ -14,22 +39,152 @@ public function _unique_element($desired = null)
...
+    protected function _add_Container($element, $args = [])
+    {
+        // Normalize the arguments, bring name out
+        if (is_string($args)) {
+
+            // passed as string
+            $args=[$args];
+        } elseif (!is_array($args)) {
+
+            throw new Exception(['Second argument must be array','arg2'=>$args]);
```
DarkSide666: why arg**2** ?

src/ContainerTrait.php
```diff
@@ -14,22 +39,152 @@ public function _unique_element($desired = null)
...
+        } elseif (is_array($args) && isset($args['name'])) {
```
DarkSide666: `is_array($args)` here is not needed anymore, because it'll always be array at this point ;)

src/ContainerTrait.php
```diff
@@ -14,22 +39,152 @@ public function _unique_element($desired = null)
...
+
+            // passed as ['name'=>'foo'];
+            $args[0]=$args['name'];
+            unset($args['name']);
+        } elseif (isset($element->short_name)) {
```
DarkSide666: Shouldn't we add `isset($element->_trackableTrait) && isset($element->short_name)` for consistancy? I mean, first check if it has trackableTrait at all and only then check for short_name property. It shouldn't change much, but I maybe that would be better if we always first check for trait existence and only then for particular property of method of this trait.

src/ContainerTrait.php
```diff
@@ -14,22 +39,152 @@ public function _unique_element($desired = null)
...
+
+            // element has a name already
+            $args[0]=$element->short_name;
```
DarkSide666: I guess should be `$args[0] = $this->_unique_element($element->short_name);` here.

src/ContainerTrait.php
```diff
@@ -14,22 +39,152 @@ public function _unique_element($desired = null)
...
+            // ask element on his preferred name, then make it unique.
+            $cn = $element->getDesiredName();
+            $args[0] = $this->_unique_element($cn);
+        } else {
+
+            // generate name based on the class
+            $cn = str_replace('\\', '_', strtolower(get_class($element)));
+            $args[0] = $this->_unique_element($cn);
+        }
+
```
DarkSide666: It looks that we have to use $this->_unique_element($desired_name) always not only in some particular cases.
That's because I can set equal names for 2 objects and then they will collide in $elements array if not making call to _unique_element().

src/ContainerTrait.php
```diff
@@ -14,22 +39,152 @@ public function _unique_element($desired = null)
...
+                'arg2'=>$args
+            ]);
+        }
+
+        if(isset($this->_appScope) && isset($element->_appScope)) {
+            $element->app = $this->app;
+        }
+
+        $element->owner = $this;
+        $element->short_name = $args[0];
```
DarkSide666: short_name property is only properly defined if $element has trackableTrait ...

src/ContainerTrait.php
```diff
@@ -14,22 +39,152 @@ public function _unique_element($desired = null)
...
-    protected function _add_Container() {
-    }
+    public function removeElement($short_name)
+    {
 
-    protected function _shorten () {
+        if (is_object($short_name)) {
+            $short_name = $short_name->short_name;
+        }
+        unset($this->elements[$short_name]);
```
DarkSide666: I remember some strange case when unset() doesn't properly free object from memory.
Shouldn't we do this better?
```
$this->elements[$short_name] = null;
unset($this->elements[$short_name]);
```

src/ContainerTrait.php
```diff
@@ -14,22 +39,152 @@ public function _unique_element($desired = null)
...
+            $len = strlen($desired);
+            $len -= ($len % $this->app->max_name_length);
+
+            $key = substr($desired, 0, $len);
+            $rest = substr($desired, $len);
+
+            if (!isset($this->app->unique_hashes[$key])) {
+                $this->app->unique_hashes[$key] = dechex(crc32($key));
+            }
+            $desired = $this->app->unique_hashes[$key].'__'.$rest;
```
DarkSide666: This way we generate names longer than max_name_length I guess.
If max_name_length is 5 and strlen($desired) = 9, then resulting string will be 14 characters long!

See full test case here: http://sandbox.onlinephpfunctions.com/code/8864da26614afdfbf0425f16039e07d67743c629

src/ContainerTrait.php
```diff
@@ -14,22 +39,152 @@ public function _unique_element($desired = null)
...
     }
 
-    function hasElement() {
+    /**
+     * Find child element by its short name. Use in chaining.
+     * Exception if not found.
+     *
+     * @param string $short_name Short name of the child element
+     *
+     * @return AbstractObject
```
DarkSide666: should be `object` instead of `AbstractObject` I guess

src/ContainerTrait.php
```diff
@@ -14,22 +39,152 @@ public function _unique_element($desired = null)
...
+
+        return $this->elements[$short_name];
     }
 
+    /**
+     * Find child element. Use in condition.
+     *
+     * @param string $short_name Short name of the child element
+     *
+     * @return AbstractObject|bool
```
DarkSide666: should be `object` instead of `AbstractObject` I guess

src/TrackableTrait.php
```diff
@@ -9,14 +9,53 @@
...
+     * in the object
+     *
+     * @var string
+     */
+    public $_trackableTrait = true;
+
+    /**
+     * Link to object into which we added this object
+     *
+     * @var AbstractObject
```
DarkSide666: should be `object` instead of `AbstractObject` I guess

src/TrackableTrait.php
```diff
@@ -9,14 +9,53 @@
...
+     *
+     * @var string
+     */
     public $short_name;
 
+    /**
+     * If name of the object is ommitted then it's natuarlly to name them
+     * after the class. You can specify a different naming pattern though
+     */
     function getDesiredName(){
```
DarkSide666: `public` function

src/TrackableTrait.php
```diff
@@ -9,14 +9,53 @@
...
-        return get_class($this);
+        return str_replace('\\', '_', strtolower(get_class($this)));
     }
 
+    /**
+     * Removes object from parent, so that PHP's Garbage Collector can
+     * dispose of it.
+     */
     function destroy(){
```
DarkSide666: `public` function

src/AppScopeTrait.php
romaninsh: Ok agree:
 - [ ] Rename all $_scopeName = true; property to use "Trait"

src/AppScopeTrait.php
romaninsh: It needs to be done on the object level, but the mechanics is only valid if the AppScope is present. If you think about it long names are only used for global scopes and long names there are not good.

Comment was copy-pasted from ATK, I haven't read it though ;) probably worth rewriting.
 - [ ] rewrite the comment for AppScopeTrait

src/AppScopeTrait.php
romaninsh: - [ ] Add description to AppScopeTrait

src/TrackableTrait.php
romaninsh: Dark, can you supply a TestCase for this?

src/ContainerTrait.php
romaninsh: - [ ] Make $desired mandatory.

src/ContainerTrait.php
romaninsh: - [ ] clean up is_array on line 56.

src/ContainerTrait.php
romaninsh: - [ ] Optimize code as requested.

src/ContainerTrait.php
romaninsh: We do not want to mess about with short_name if it's set. 

I have to think of the implications here, because you can add object in multiple locations in theory. Let's leave it as-is for now until we have some realistic test-cases or use-cases.

src/ContainerTrait.php
romaninsh: If names collide we create exception. We do not want field names to be renamed. So we only generate unique names if they are auto-generated from class name or from getDesiredName()

src/ContainerTrait.php
romaninsh: Yes, i thought about that too. but sometimes we want to add object which will not be tracked.

In this cases - what do we actually do? there is no point calling this method at all.  So I returned those lines. 

Maybe if element is not trackable and it is passed to _add_Container we should just spit it out right away?

The other option is that the element still wants some unique identifier, which is what we actually do here and ATK would want it like this. The alternative is to add UnTrackableTrait. Let's leave until we have more suggestions here.

src/ContainerTrait.php
romaninsh: I think unset is sufficient. It's what we currently have in ATK.

src/ContainerTrait.php
DarkSide666: This probably is related with this comment too: https://github.com/atk4/core/pull/3#discussion_r63666353

src/ContainerTrait.php
romaninsh: I have refactored the logic of the method. It now uses chunks of max_name_length to cut. Assume max-name_length = 10. And length of the string is 34.

The $len will be 30 here, so it will cut off 30 and convert that into the key. The rest 4 characters will remain. 

True that after re-attaching the hash on the front it will be more than 10, if we increase len, we should check to make sure we don't leave empty space at the end. 

This needs more test-cases.

The reason why i'm evening the cuts is to avoid situation like this:

 - 281937_bla_hello_world
 - 199402_la_hello_world2

This works with the current ATK cutting too, but for situations when short_name is much longer - it won't cut off enough.

 - [ ] Add more test-scripts an change this method if necessary.

src/ContainerTrait.php
romaninsh: Copy-paste 👍 

 - [ ] replace AbstractObject with atk4\core\Object

src/ContainerTrait.php
romaninsh: Trackable trait.

src/TrackableTrait.php
romaninsh: i always forget these ;)

src/TrackableTrait.php
romaninsh: Addressed.

romaninsh:
All comments have been addressed.

src/ContainerTrait.php
```diff
@@ -2,34 +2,213 @@
...
+     * If you are using ContainerTrait only, then you can safely
+     * use this add() method. If you are also using factory, or
+     * initializer then redefine add() and call
+     * _add_Container, _add_Factory, 
+     */
+    function add($obj, $args = [])
+    {
+        if (isset($this->_factoryTrait)) {
+            // Factory allows us to pass string-type objects
+            $obj = $this->_add_Factory($obj, $args);
```
skondakov: Aren't we following some king of CS for methods naming i.e. shouldn't $this->_add_Factory be $this->_addFactory?

romaninsh: yes, but ContainerTrait also adds add() method which can integrate nicely with others. I tried to keep add() method outside of traits and it proved to be a pain.

src/ContainerTrait.php
```diff
@@ -2,34 +2,213 @@
...
     /**
-     * Extension to add() method which will perform container tracking
+     * Extension to add() method which will perform linking of
+     * the object with the current class.
      */
-    protected function _add_Container() {
+    protected function _add_Container($element, $args = [])
+    {
+        if (!is_object($element)) {
+            throw new Exception(['Only objects may be added into containers','arg'=>$element]);
```
skondakov: Shouldn't we check here if added object is a container as well?

romaninsh: You mean element? Well it actually dosen't have to be a container, there are test-script which checks it: https://github.com/atk4/core/pull/3/files#diff-afb03a24bbd48ed1eff2eb5456dc0b24R22

PR #5, opened 2016-05-21
Feature/implement initializer

romaninsh:
Initializer trait implemented.

This will add ability for objects to execute init() method. Docs are also updated.

Merge this ONLY after #3 is merged (because this PR is based on it's branch)


skondakov:
Seems good to me! Great work!

PR #6, opened 2016-05-21
Feature/implement hooks

romaninsh:
Hooks have been implemented.

This will add ability for objects to use `addHook()` and `hook()`. Includes docs.

Merge this ONLY after #3 is merged (because this PR is based on it's branch)


src/HookTrait.php
```diff
@@ -4,17 +4,165 @@
...
+
+        // Set defaults
+        if (is_null($arguments)) {
+            $arguments = [];
+        } elseif (!is_array($arguments)) {
+            throw new Exception(['Incorrect arguments for addHook','args'=>$arguments]);
+        }
+        if (is_null($priority)) {
+            $priority = 5;
+        }
```
skondakov: You can set a default value for priority of 5 and not check if is null here

tests/HookTraitTest.php
```diff
@@ -0,0 +1,184 @@
...
+}
+
+class HookMock {
+    use HookTrait;
+
+    public $result = 0;
+
+    function myCallback($obj) {
+        $this->result++;
+    }
```
skondakov: Very good & detailed UTs ;)

skondakov:
Seems good to me!

PR #7, opened 2016-05-21
Feature/implement dynamic methods

romaninsh:
Implemented dynamic methods

    $obj->addMethod('test', callback)

    $obj->test()

### DIFF
--- a/src/DynamicMethodTrait.php
+++ b/src/DynamicMethodTrait.php
@@ -76,11 +76,11 @@ trait DynamicMethodTrait {
     public function hasMethod($name)
     {
         return method_exists($this, $name)
-            || $this->hasHook('method-'.$name)
+            || $this->hookHasCallbacks('method-'.$name)
             || (
                 isset($this->_appScopeTrait) &&
                 isset($this->app->_hookTrait) &&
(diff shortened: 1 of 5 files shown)


src/DynamicMethodTrait.php
```diff
@@ -4,20 +4,147 @@
 
 trait DynamicMethodTrait {
 
-    function __call($fx, $args){
+    public function __call($method, $arguments)
+    {
+        if (($ret = $this->tryCall($method, $arguments))) {
```
skondakov: One brackets here seem not needed -> if ($ret = $this->tryCall($method, $arguments)) {

romaninsh: It's a tricky one. Basically some people may see this as mistake where `==` should be used instead of `=` , thats why (( )) are used. It's some sort of tradition from C days. I'll try to find reference.

skondakov: I see, ok then leave it as is.

skondakov:
Seems it is good to go as well.

src/DynamicMethodTrait.php
romaninsh: http://stackoverflow.com/a/15379900/204819

PR #8, opened 2016-05-28
Add testing of reference passing

romaninsh:
This adds test that would allow passing references into a hook:

    $o->addHook('inc', function(&$a){ $a++; });

    $arg = 1;
    $o->hook('normalize', [&$arg]);
    echo $arg;  // 2


skondakov:
Looks good to me!

PR #9, opened 2016-06-20
Implementing colorful exception

romaninsh:
```
echo $exception->getColorfulText();
```

when using the console. Will result in a nicely looking colorful exception.

screen shot 2016-06-19 at 15 04 40


skondakov:
A good way to show console error ;) No questions here as well!

PR #13, opened 2016-07-12
clenaup implementation

romaninsh:


DarkSide666:
All looks fine except that variable `$postfix` should be better to name `$prefix` :)

PR #23, opened 2016-07-18
Implement session trait

DarkSide666:
Simple implementation of SessionTrait.

### DIFF
--- a/docs/session.rst
+++ b/docs/session.rst
@@ -12,15 +12,33 @@ Introduction
 Properties
 ==========
 
-    None.
+.. php:attr:: session_key
+
+    Internal property to make sure that all session data will be stored in one "container" (array key).
 
 Methods
 =======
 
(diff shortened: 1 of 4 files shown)


romaninsh:
@DarkSide666 to answer your question - different frameworks implement session in various ways. The point of Agile Core / Data and DSQL is to work anywhere.

Because core functionality of DSQL / Data do not require use of sessions, we shouldn't even implement them. I'm merging the request, but it will be migrated into the ATK framework when it's ready.

romaninsh:
Actually we should probably put warning about this, we don't want users to start rely on Session and then remove them. 

It's a head-scratcher ;)

DarkSide666:
well, maybe you shouldn't merge sessions then and simply delete that branch?

romaninsh:
I'll add warning into documentation.

DarkSide666:
ok

PR #33, opened 2017-01-26
Some improvements for better integration with Agile UI

romaninsh:
- improve handling of short-names (bugfixes)
 - delegate 'normalizeClassName' to the $app if it is set and method exists.

This would allow App to set default namespaces and class routes, e.g.:

    $this->add('Grid');

instead of

    $this->add('atk4/ui/Grid');

### DIFF
--- a/src/FactoryTrait.php
+++ b/src/FactoryTrait.php
@@ -33,8 +33,6 @@ trait FactoryTrait
             ]);
         }
 
-        $object = $this->normalizeClassName($object);
-
         return new $object($defaults);
     }
 
--- a/src/FactoryTrait.php
+++ b/src/FactoryTrait.php
@@ -49,7 +49,7 @@ trait FactoryTrait
         }
 
         $object = $this->normalizeClassName($object);
-        
+
         return new $object($defaults);
     }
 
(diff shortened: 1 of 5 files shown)


src/FactoryTrait.php
```diff
@@ -33,8 +48,6 @@ public function factory($object, $defaults = [])
             ]);
         }
 
-        $object = $this->normalizeClassName($object);
```
DarkSide666: why you took this out? from where do we call normalizeClassName method then?

src/FactoryTrait.php
```diff
@@ -54,10 +67,26 @@ public function factory($object, $defaults = [])
      */
     public function normalizeClassName($name, $prefix = null)
     {
+        if (is_array($name) && isset($name[0])) {
```
DarkSide666: can you give example how this part can be useful?

src/FactoryTrait.php
```diff
@@ -54,10 +67,26 @@ public function factory($object, $defaults = [])
...
+            $name[0] = $this->normalizeClassName($name[0]);
+
+            return $name;
+        }
+
         if (!is_string($name)) {
             return $name;
         }
 
+        if (isset($this->_appScopeTrait) && $this->app) {
```
DarkSide666: this can be moved slightly up and can use already existing `if (!is_string($name)) return $name;` part above.

DarkSide666:
Also Travis test fails:
```
1) atk4\core\tests\FactoryTraitTest::testException1
Failed asserting that exception of type "Error" matches expected exception "Exception".
```

Issue #46, opened 2017-04-12
Undefined factory defaults shouldn't throw exception.


romaninsh:
Calling factory like this:

```
$m1 = $m->factory('atk4\core\tests\FactoryMock', ['not_exist'=>'test']);
```

may encounter some parameters that are not defined in the class. Instead of throwing exception we are going to silently ignore them for a few reasons:

 - Agile UI redefines this to bring class support
 - 'desired_name' is used for name hinting and it's not a real property
 - component may remove support for feature which shouldn't trigger exception
 
So we are disabling this check.

romaninsh:
Implemented in https://github.com/atk4/core/commit/a1e3f343d319e2b53b25518f43a019329bd04ce2

PR #47, opened 2017-04-12
Implement DI Container Trait for Defaults

romaninsh:
Related to #46..

### DIFF
--- a/src/FactoryTrait.php
+++ b/src/FactoryTrait.php
@@ -28,26 +28,11 @@ trait FactoryTrait
 
         if (is_object($object)) {
 
-            // If some defaults are passed, set them up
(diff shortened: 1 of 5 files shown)


src/FactoryTrait.php
```diff
@@ -30,8 +30,8 @@ public function factory($object, $defaults = [])
 
             // If object implements DIContainerTrait we can inject some
             // of the properties without causing harm
-            if (is_array($defaults) && isset($object->_DIContainerTrait)) {
```
romaninsh: are you sure? there may be some garbage there..

DarkSide666: DI partly takes care of that, but you're right - will rollback this

PR #53, opened 2017-09-03
Improve DebugTrait

romaninsh:


src/DebugTrait.php
```diff
@@ -14,74 +14,121 @@
...
+                $bt[] = $line['file'].':'.$line['line'];
             }
+        }
 
-            $this->_prev_bt[$trace] = $bt;
+        if (isset($this->_prev_bt[$trace]) && array_diff($this->_prev_bt[$trace], $bt)) {
+            $d1 = array_diff($this->_prev_bt[$trace], $bt);
+            $d2 = array_diff($bt, $this->_prev_bt[$trace]);
+
+            $this->log('debug', 'Call path for '.$trace.' has diverged (was '.implode(', ', $d1).', now '.implode(', ', $d2).")\n");
```
DarkSide666: Shouldn't this line better be like below ?
```
$this->debug(true); // force enable debug mode
$this->debug('Call path for '.$trace.' has diverged (was '.implode(', ', $d1).', now '.implode(', ', $d2).")"); // no \n at the end
```

romaninsh: log('debug') will work same as debug(), so this way it's simpler.

PR #59, opened 2017-09-12
setDefaults should ignore numeric keys otherwise we can get exception…

DarkSide666:
setDefaults should ignore numeric keys otherwise we can get exceptions like these:

```
atk4\core\Exception: Property for specified object is not defined
             object: Object atk4\data\Model
           property: 0
              value: "user"
```


romaninsh:
Actually setDefaults shouldn't be passed in any numeric values. Cal you tell me who did that?

PR #66, opened 2017-11-19
Add SQLTestCase

romaninsh:
Agile Data implements a local class called "SQLTestCase", that implements few very useful routines:

 - setDB([]) - initialize database tables from array
 - getDB() - get database configuration

The application in test suites is very useful. 

``` php
        $this->setDB($q = ['user' => [
            ['name' => 'John', 'surname' => 'Smith'],
            ['name' => 'Steve', 'surname' => 'Jobs'],
        ]]);

        // do something with database

        $state = $this->getDB('user');
```

### DIFF
--- a/src/PHPUnit_SQLTestCase.php
+++ b/src/PHPUnit_SQLTestCase.php
@@ -1,0 +1,142 @@
+<?php
+
+namespace atk4\core;
+
+use atk4\data\Persistence_SQL;
+use atk4\dsql\Query;
+
+class PHPUnit_SQLTestCase extends PHPUnit_AgileTestCase
+{
+    public $db;
+
+    public $tables = null;
+
+    public $debug = false;
+
+    public function setUp()
+    {
+        parent::setUp();
+        // establish connection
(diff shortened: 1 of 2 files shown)


src/PHPUnit_SQLTestCase.php
```diff
@@ -0,0 +1,142 @@
...
+    /**
+     * Sets database into a specific test.
+     */
+    public function setDB($db_data)
+    {
+        $this->tables = array_keys($db_data);
+
+        // create databases
+        foreach ($db_data as $table => $data) {
+            $s = new Structure(['connection' => $this->db->connection]);
```
romaninsh: TODO: replace with atk4/schema/Migration

src/PHPUnit_SQLTestCase.php
```diff
@@ -0,0 +1,142 @@
...
+                }
+            }
+
+            $ret[$table] = $data2;
+        }
+
+        return $ret;
+    }
+
+    public function runBare()
```
romaninsh: i think this method already exists in Agile Test Case.

romaninsh:
I suspect this might be a duplicate

Issue #65, opened 2017-10-30
MergeSeeds don't respect numeric keys defined as strings


DarkSide666:
Example:
```
$years = ['2013', '2014'];
$a = new Object(['years'=>$years]);
var_dump($a->years); // returns [0=>'2013', 1=>'2014']
```

That's because `is_numeric('123')===true`. Must use `is_int('123')===false` almost everywhere.

In here:
https://github.com/atk4/core/blob/develop/src/FactoryTrait.php#L38
https://github.com/atk4/core/blob/develop/src/FactoryTrait.php#L59

Also `is_numeric` is quite widely used in DIContainer and in data, dsql and ui repositories. Need to review them too.

romaninsh:
Unable to reproduce. Tried:

```
        $s1 = $this->factory(new SeedDITestMock(), ['foo'=>['bar', 'baz']]);
        $this->assertEquals(['bar', 'baz'], $s1->foo);
```

romaninsh:
reviewing the example you have provided, it seems that it's correct. `[0=>'2013', 1=>'2014']` is identical to `['2013', '2014']` isn't it?

DarkSide666:
I mean like this:

```
$years = ['2013'=>'2013', '2014'=>'2014'];
$a = new Object(['years'=>$years]);
var_dump($a->years); // returns [0=>'2013', 1=>'2014']
```

romaninsh:
@DarkSide666 is this fixed ?

DarkSide666:
Probably I was talking about these lines `array_filter($seed2, 'is_numeric', ARRAY_FILTER_USE_KEY); // with numeric keys`, but not sure what exactly was the problem.

I think it's good as it is now if nobody else have this problem for years.

PR #67, opened 2017-11-29
Feature/no special treatment for injected arrays

romaninsh:
When DIContainerTrait was originally implemented, it had a special treatment for arrays, e.g:

``` php
$app->add('Button', 'class'=>['blue']);
```

Instead of replacing value of the $class property, the contents would be merged. Unfortunately this has a major impact on the way how "seeds" work, for example here:

``` php
$app->add('MyWidget', 'title'=>['MyTitle', 'hello world']);
```

If MyWidget already defines title property it won't be replaced with new seed instead it will be merged causing 'MyTitle' to be sent as constructor argument.

There are many cases where array merging is bad and only a single case where it's beneficial, so I'm making array injection consistent with other values and will provide a special treatment for the $class property merging inside views.

I have supplied a test-script which can be used to refactor View's `setDefaults` method in a test-script within this PR

 - [ ] Check if any documentation should be removed / amended.

### DIFF
--- a/tests/SeedTest.php
+++ b/tests/SeedTest.php
@@ -128,7 +128,7 @@ class SeedTest extends \PHPUnit_Framework_TestCase
     public function testMerge4()
     {
         // array values don't overwrite but rather merge
-        $o = new SeedDITestMock();
+        $o = new ViewTestMock();
         $o->foo = ['red'];
         $oo = $this->mergeSeeds(['foo'=>['green']], $o);
 
@@ -147,9 +147,9 @@ class SeedTest extends \PHPUnit_Framework_TestCase
     public function testMerge5()
     {
         // works even if more arguments present
-        $o = new SeedDITestMock();
+        $o = new ViewTestMock();
         $o->foo = ['red'];
-        $oo = $this->mergeSeeds(['foo'=>['xx']], ['foo'=>['green']], $o, ['foo'=>5]);
+        $oo = $this->mergeSeeds(['foo'=>['xx']], ['foo'=>['green']], $o);
 
         $this->assertSame($o, $oo);
         $this->assertEquals($oo->foo, ['red', 'green', 'xx']);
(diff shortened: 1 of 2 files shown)


tests/SeedTest.php
```diff
@@ -161,16 +161,37 @@ public function testMerge5()
...
-        $this->assertEquals($oo->foo, ['red', 'xx']);
+        $this->assertEquals(['red', 'xx'], $oo->foo);
+    }
+
+    public function testMerge6()
+    {
+        $oo = $this->mergeSeeds(['4'=>'four'], ['5'=>'five']);
+        $this->assertEquals($oo, ['4'=>'four', '5'=>'five']);
+
+        $oo = $this->mergeSeeds(['4'=>['four']], ['5'=>['five']]);
```
DarkSide666: this test is duplicated above

tests/SeedTest.php
```diff
@@ -161,16 +161,37 @@ public function testMerge5()
...
+        $this->assertEquals($oo, ['4'=>['four'], '5'=>['five']]);
+
+        $oo = $this->mergeSeeds(['x'=>['four']], ['x'=>['five']]);
+        $this->assertEquals($oo, ['x'=>['four']]);
+
+        $oo = $this->mergeSeeds(['4'=>['four']], ['4'=>['five']]);
```
DarkSide666: I wonder why these don't fail, because we still have that `is_numeric` / `is_int` issue: https://github.com/atk4/core/issues/65 not fixed, so when merging these keys `'4'` it should merge these values.

romaninsh: hehe me neitherr. in fact i added 'invalid' to that ticket, so when you have time create a proper case for it.

tests/SeedTest.php
romaninsh: no, the value is array, above value is string.

PR #73, opened 2018-01-12
Feature/php72 no longer has stderr

romaninsh:


DarkSide666:
PHP 7.1.9 which i use also don't have STDERR constant.

PR #75, opened 2018-02-08
numeric key seed always should go trough

### LINKED ISSUE #74: Injecting "null" argument into constructor fails
Suppose you create a button using a seed:

``` php
$button = $app->add(['Button', 'icon'=>'book']);
```

it is created with no label and with an icon. Next consider this code:

``` php
$button = $app->add(['Button', 'lablabla', 'icon'=>'book']);
```

This works as expected. Button with label and icon is added. Next try this:

``` php
$button = $app->add(['Button', 'lablabla', 'big red', 'icon'=>'book']);
```

once again, this works as expected. A big red button with label and icon is added.

Now suppose you want to get rid of label, so you replace it with `null`

``` php
$button = $app->add(['Button', null, 'big red', 'icon'=>'book']);
```

But instead you find with a small button and label "big red". 

A current workaround is to specify 2nd argument as "false" or empty string but I think `null` should be equally acceptable.

DarkSide666:
fix #74

### DIFF
--- a/src/FactoryTrait.php
+++ b/src/FactoryTrait.php
@@ -26,7 +26,6 @@ trait FactoryTrait
      */
     public function mergeSeeds($seed, $seed2, ...$more_seeds)
     {
-
         // recursively merge extra seeds
         if ($more_seeds) {
             $seed2 = $this->mergeSeeds($seed2, ...$more_seeds);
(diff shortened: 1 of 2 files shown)


romaninsh:
i'll test this in my project..

PR #77, opened 2018-03-05
implement PSR3 support in DebugTrait

romaninsh:
add debug trait, then call $this->info or $this->critical.....


romaninsh:
documentation is somewhat innacurate but needs a proper rewrite..

romaninsh:
@DarkSide666 it's like we agreed on this implementation.

PR #88, opened 2019-03-11
Fix multiple files readConfig

abbadon1334:
if more than one file is read $config will store all files merged config not only last loaded

### DIFF
--- a/src/ConfigTrait.php
+++ b/src/ConfigTrait.php
@@ -44,8 +44,6 @@ trait ConfigTrait
      */
     public function readConfig($files = ['config.php'], $format = 'php')
     {
-        $this->config = [];
-
         if (!is_array($files)) {
             $files = [$files];
         }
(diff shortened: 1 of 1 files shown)


src/ConfigTrait.php
```diff
@@ -54,33 +52,37 @@ public function readConfig($files = ['config.php'], $format = 'php')
...
                     if (!class_exists(\Symfony\Component\Yaml\Yaml::class)) {
                         throw new Exception(['You need Symfony\Yaml repository if you want to parse YAML files']);
                     }
-                    $this->config = \Symfony\Component\Yaml\Yaml::parseFile($file);
+                    $tempConfig = \Symfony\Component\Yaml\Yaml::parseFile($file);
                     // @codeCoverageIgnoreEnd
                     break;
             }
+            
+            $this->config = array_merge($this->config, $tempConfig);
```
DarkSide666: `array_merge` or `array_merge_recursive` here?
keep in mind that config options can have deep structure too so i think array_merge_recursive would be better.

src/ConfigTrait.php
```diff
@@ -44,8 +44,6 @@ trait ConfigTrait
      */
     public function readConfig($files = ['config.php'], $format = 'php')
     {
-        $this->config = [];
```
DarkSide666: If you remove this line, then you have to make sure $this->config is empty array at the beginning. So maybe let's set it as empty array in class properties - line 29.
or in line 83 do like this:
`$this->config = array_merge_recursive($this->config ?: [], $tempConfig);`

PR #93, opened 2019-06-09
Wip translation

abbadon1334:
work in progress, but already useable without using gettext because for what i have seen is used as a singletone at low level and can't be changed ( or i didn't find any prove i was wrong about this).

I still think about last discussions about this topic and we speak about having App has a reference for object translator, but after that we speak about addons that can have some defined translation even in saasty, this translations can impact the standard translations that we have in ui or another addon, 3 possible solution : 

- To avoid this why not create a trait TranslatorTrait and the class that use this trait in place of passing the app->translator reference it will pass to his childs her ->translator reference and act as a first try to translate and if not present fallback to App->translator?
- Every addon must have a uuid and it will be in front of every string to be translated
- I already use ConfigTrait in Translator class, we can added addons translation to the config, but if will be add a poor translation it will break some of the others translations.

Choose your destiny :D  and give feedback on this wip PR


src/TranslatableTrait.php
```diff
@@ -0,0 +1,74 @@
...
+            $sub_args = $sub_args[1] ?? false;
+
+            if (!$sub_args) {
+                $translated_args[] = $this->__($sub_string);
+            } else {
+                $translated_args[] = $this->__($sub_string, $sub_args);
+            }
+        }
+
+        return call_user_func_array('sprintf', $translated_args);
```
romaninsh: 1. use TransatableTrait
2. $this->__('we add %s into %s', ['milk', 'bowl']);

^^ the only thing trait should do is check for API and fallback to sprintf..

src/TranslatableTrait.php
```diff
@@ -0,0 +1,74 @@
...
+     * Translate a string.
+     *
+     * @param string $string
+     * @param mixed  ...$args
+     *
+     * @throws Exception
+     *
+     * @return string
+     */
+    public function __(string $string, ...$args)
```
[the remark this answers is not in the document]
abbadon1334: i change it with the new implementation defined in meeting 13-06

tests/TranslatableMock.php
```diff
@@ -0,0 +1,38 @@
...
+{
+    use TranslatableTrait;
+
+    private $translation = [
+        'string without counter'       => 'string without counter translated',
+        'string not translated simple' => [
+            1 => 'string translated',
+        ],
+        'string not translated with plurals' => [
+            0 => 'string translated zero',
```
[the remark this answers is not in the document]
abbadon1334: i will do :D

abbadon1334:
Simplify everything.
- removed sprintf for recursive translations, sprintf if needed is done directly in code (suggested by pkly) 
- add context of translations with default = atk4
- simplify interface
- removed final from translator class, that will be moved in a separated addon and added to composer suggest 

- in App must be implemented only method : 
```php
_(string $message, int $count = 1, string $context = 'atk4'): string
```
- every class that use the trait for easy usage, have 2 methods

```php
public function _(string $message, int $count = 1, string $context = 'atk4'): string;
public function _d(string $message, string $context = 'atk4', int $count = 1): string;
```
just because the first can be used like this : 
```php
$this->_('apple'); // => mela ( singular )
// OR
$this->_('apple',2); // => mele ( plural 2 form )
// OR
$this->_('apple',0); // => nessuna mela ( zero form)

// WITH DOMAINS

$this->_d('apple', 'my_domain'); // => mela ( singular )
// OR
$this->_d('apple', 'my_domain', 2); // => mele ( plural 2 form )
// OR
$this->_d('apple', 'my_domain', 0); // => nessuna mela ( zero form)

// every call above will call the method if exists
$this->app->_(string $message, int $count = 1, string $context = 'atk4'): string
```
i choose to split the method in two for easy usage because we need to avoid the check for numeric or string or count of parameters, in this way even autocompletation works correctly, 

TODO : added test for context domain
TODO : added more test in general

@romaninsh, @DarkSide666 what we do in the app for adding the method _?
 - did you think we need a TranslatorTrait for the app?
 - we add the method _ directly in \atk4\ui\App ?

i think the second is correct, but i have an idea : 
If we make a TranslatorTrait for the object that produce translations; In place of looking to the app, we can look to the first ->owner that use TranslatorTrait, in this way we can propagate from parent to every child the translations and make something really interesting.

src/Translator.php
```diff
@@ -0,0 +1,152 @@
...
+        $trans = $this->translation[$context][$message] ?? false;
+
+        // if not present return string
+        if (false === $trans) {
+            return $message;
+        }
+
+        // if present but empty raise exception
+        if (empty($trans)) {
+            throw new Exception('Translation is present but is empty');
```
pkly: I'd suggest returning a default string instead of an exception here, possibly just a static one eg. "No translation found", or the original text with that string added to it.

pkly:
I wanted to propose so that the App controls all translations, but that doesn't seem like the greatest idea if you'd ever want to translate anything that's not using the Atk4/ui components and the App class.

I thought that we could just add the translation trait to App and View, and have the other elements simply use a simplified trait, that'd try to pass the translation to either App (via AppScope) or View (via TrackableTrait -> owner?) and run the translation there.

That way all classes (outside of Atk4/ui) would not "support" translation, but instead have a dummy trait that'd specify all the required functions, and when used together with UI, they would have access to the translation class governed by App/View.

How does that sound?

src/Translator.php
abbadon1334: is only an Exception that raise only for integrity when a translation was defined in a bad way.
Translation will be defined like this
```php
$translation['domain'] = [
    'stringA' => [
        0 => 'no string',
        1 => 'string',
        2 => 'strings',
    ]
];
```
It will raise only if defined like this :
```php
$translation['domain'] = [
    'stringA' => []
];
```

src/Translator.php
```diff
@@ -0,0 +1,152 @@
...
+    {
+
+        // check presence of key
+        $trans = $this->translation[$context][$message] ?? false;
+
+        // if not present return string
+        if (false === $trans) {
+            return $message;
+        }
+
```
abbadon1334: i must added a check for !is_array($trans) return $message, in case of declaration like this
```php
$translation['domain'] = [
    'stringA' => 'no string'
];
```

src/TranslatorInterface.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+namespace atk4\core;
+
+interface TranslatorInterface
```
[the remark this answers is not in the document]
abbadon1334: the only one i found is this : https://gist.github.com/oscarotero/33e3af8741045c2a1a5a89310571cbdb

abbadon1334:
> I wanted to propose so that the App controls all translations, but that doesn't seem like the greatest idea if you'd ever want to translate anything that's not using the Atk4/ui components and the App class.
> 
> I thought that we could just add the translation trait to App and View, and have the other elements simply use a simplified trait, that'd try to pass the translation to either App (via AppScope) or View (via TrackableTrait -> owner?) and run the translation there.
> 
> That way all classes (outside of Atk4/ui) would not "support" translation, but instead have a dummy trait that'd specify all the required functions, and when used together with UI, they would have access to the translation class governed by App/View.
> 
> How does that sound?

If you use singleton you can use it even outside of atk4, if you want to integrate in atk4 you can do it in 3 way : 
- use AppScope/Translatable Traits
- create a composite pattern that inject translator in your classes
- expose the translator as a singleton and use it out of app

i already put out an idea in the previous message :
> If we make a TranslatorTrait for the object that produce translations; In place of looking to the app, we can look to the first ->owner that use TranslatorTrait, in this way we can propagate from parent to every child the translations and make something really interesting.

in this way we can avoid the use of AppScope in library out of atk and use : 
TrackableTrait+TranslatorTrait => for the class that can translate to children
TrackableTrait+TranslatableTrait => for the class that needs translations

@pkly If you want to contribute even with a total different implementation, you are welcome! but i think the problem is not how to use it in atk than using it with minimum changes out of atk4 without confusing the developer

pkly:
Ah, right, my bad.
Yeah, TrackableTrait + TranslateableTrait seems like a good idea.

As for how useful it'd be outside of atk4 and everything related to that - can't say I really have an idea there, as I'm already using a custom solution that's just a singleton, mostly because I need it to be bound with the CMS I'm building on top of atk.

src/Translator.php
pkly: Still feels wrong for me to print out exceptions in a translation class, since it can break otherwise valid code for a reason that's only partially worth it.
I'd suggest creating an error message instead and attaching it to the front of the translation (thus somewhat breaking it), but making it still useable in the worst-case scenario.

abbadon1334:
@DarkSide666 said something really interesting, think about a page with two view with different languages without using two app, that can be awesome in many situation like managing languages in a cms or manage template emails, not needed OK, but in some specific cases can be a saver. Using a singleton make everything much easier, but after you enter that path you cannot return and make everything too much coupled.

pkly:
Well yeah, I do understand the use scenario.
My point is that it's questionably useful outside of atk. I do think that's the best way for atk to implement it, but it should be up to the end-user if they want to expand on that by using it in their projects as well.

I already went the singleton route for my translation needs, I needed to have every string possible already translated from the get-go, because I can't release a product that doesn't support unlimited UI languages.

Talking about languages and all, if you'd like I could give you the code for the translation engine I've already built - you could pull out the functions for searching files and such, it allows to load translated strings from .json files (lazily), so it's pretty comfortable to use. You'd just need to modify it for the numbered scenarios.

abbadon1334:
> Talking about languages and all, if you'd like I could give you the code for the translation engine I've already built - you could pull out the functions for searching files and such, it allows to load translated strings from .json files (lazily), so it's pretty comfortable to use. You'd just need to modify it for the numbered scenarios.

Sure can be inspiring because you probably have already faced some problem and found a solution.

pkly:
Here's the code for it https://pastebin.com/nf4jx8ci
A few notes: 

- This assumes all translations are in json files in pairs "Original English Text": "Translated Text"
- It also tries to load translations by default on the name of a class, in your case you could swap it for the domain (/i18n/atk/en.json | /i18n/atk/it.json) etc.
- It loads the whole file if the translation is requested simply because it's very likely that the file will be required anyway, and constantly seeking for translations / opening and closing files seems like a waste of time.
- It has some code (determineLanguage mostly) that's related to my CMS, but you can somewhat ignore that.
- It assumes you allow languages to be used eg.

```php
StaticTranslation::allowLanguage('pl');
StaticTranslation::allowLanguage('en');
```

This is because I will not allow anything that's not passed through some QC at least and deemed useable, no need to allow all languages anyway when there's no translations available.

abbadon1334:
i see the code and i already do what you do there, i try to added more comment in the next commit to show it better the reasoning behind every operations in Translator => even if it will be moved as an addon, to be much more understandable.

src/Translator.php
romaninsh: I vote to define this through a parameter.

tests/TranslatableTraitTest.php
```diff
@@ -0,0 +1,63 @@
...
+
+    public function testSingularNotExists()
+    {
+        $trans = $this->translatableMock->_('string not exists');
+        $this->assertEquals('string not exists', $trans);
+    }
+
+    public function testPlurals0()
+    {
+        $trans = $this->translatableMock->_('string not translated with plurals', 0);
```
romaninsh: Please use more meaningful examples. I still don’t understand singular/plural logic. Also what if it uses multiple arguments: (3 record(s) imported. 1 error(s))

src/Translator.php
abbadon1334: like we do in App, is the best option, but in case of not raising exception, even if we return string untranslated or empty, we leave the user with no clue of where can be the error.

add flag $raise_bad_format_exception = boolean;
what returns? and empty string or the untranslated string?

romaninsh: I’d log that and then add all translations automatically

pkly: I'd suggest retuning the untranslated string.

romaninsh:
Be sure to have a simple and usable solution which can be extended. In case Dark suggested with multiple pages - it’s not practical and even if needed - app can check property of a page. We can make trait pass original object to app so that we know who caller was.

romaninsh:
Object without an app again is not practical. If trait does all work it cannot be affected by dev.

pkly:
@romaninsh if we'd like to use the suggested way of doing

> in this way we can avoid the use of AppScope in library out of atk and use :
TrackableTrait+TranslatorTrait => for the class that can translate to children
TrackableTrait+TranslatableTrait => for the class that needs translations

Wouldn't it simply be possible to put the TranslateableTrait into all classes that have any strings that need translations, and then wrap them at any point into a container that'd use the TranslatorTrait and override the methods?
That way we could keep the translation solution exposed to the outside, although I'm unsure how you'd cause the translation to run let's say in core itself.
I'm unsure if core contains somekind of endpoint or wrapping class, that can be used to implement the TranslatorTrait.

Otherwise in atk/ui we can simply put the TranslatorTrait in the App, since that you can simply override and it's not a problem.

I think the methods themselves, along with the basic implementation can simply be hardcoded in atk/core (TranslatableTrait + TranslatorTrait), since the translateable functionality is just a single function call and will not differ in different implementations.

If need be, one can simply override the class used for displaying the end result instead of modifying the trait directly, since you can just override the methods declared in the TranslatorTrait.

abbadon1334:
adding context domain is a pain in the a... but i solved.
@romaninsh test are better now? if not can we speak together and see how i can improve that part?
@pkly there is one single problem for TranslatorTrait|TranslatableTrait, the chain parent=>children must be contigous to work
@DarkSide666 at now it checks only for presence of 
```php
$this->app->_($string,$count,$context); 
```
we can add, and extra check for $this->translator and call it directly like in the app
```php
$this->translator->translate(string $string, ?int $count = 1, ?string $context = NULL);
```
if app and translator are not present return string

i think this way we can use it wherever we want and inject translator where is needed out of atk4.

I use the ConfigTrait in Translator class, to loading the config but after every call i reset $this->config to  not reinvent the wheel, i have to do this, because array_merge_recursive add values in place of replacing it, i use array_replace_recursive to load the fallback language and overwrite values with the requested language.

abbadon1334:
i know i need to add more comments on some methods , but i want to make more "clear" the direction of the progress

src/Translator.php
```diff
@@ -0,0 +1,158 @@
...
+     * @param string      $language
+     *
+     * @param string|null $fallback
+     *
+     * @return string
+     */
+    public function setLanguage(string $language,?string $fallback = null)
+    {
+        $this->language = $language;
+        $this->fallback = $fallback ?? false;
```
pkly: Maybe change this to 
```php
$this->fallback = $fallback ?? $this->fallback ?? false;
```

src/Translator.php
```diff
@@ -0,0 +1,158 @@
...
+     * @var string
+     */
+    public $language;
+
+    /**
+     * ISOCode of the fallback language.
+     *
+     * @var string
+     */
+    public $fallback;
```
pkly: Set default value to 'en', since that's what all the strings are in anyway, so it should be the fallback by default.

src/Translator.php
abbadon1334: correct and linear, but we already return the string which by default is in english. but i think this must be done
@romaninsh @DarkSide666 what do you think?

abbadon1334: cannot be false if we make $this->fallback=en

need to add an exception in setLanguage or move the set of language and fallback directly in costructor to avoid throwing, because language must be defined only ones.

src/TranslatableTrait.php
abbadon1334: do it but i add two helper method that make the usage of the trait much short and declarative with array. look in this test for helper usage :
https://github.com/abbadon1334/core/blob/dc07b17f949fdea4f42c7b2b56351b5dd15736e1/tests/TranslatableTraitAppTest.php#L271

src/Translator.php
abbadon1334: add flag : https://github.com/abbadon1334/core/blob/dc07b17f949fdea4f42c7b2b56351b5dd15736e1/src/Translator.php#L35

tests/TranslatableTraitTest.php
abbadon1334: Change some naming and added multiple arguments, see if is better now

src/Translator.php
pkly: Then just don't allow false as a value for fallback.
You've already listed the type as string, and it makes way more sense to simply have it as 'en' rather than false.
There's always going to be some text, in the worst case scenario you can simply return the original string.

No need for an exception, I think, but you could be missing some code that 'allows' languages.
Since now, there's no way to know what is valid and what is not, so you must assume that any string is a valid language.

abbadon1334:
last push for today.

src/Translator.php
```diff
@@ -38,10 +38,10 @@ class Translator implements TranslatorInterface
      * @param string      $language Primary Language
      * @param string|null $fallback Fallback Language
      */
-    public function __construct(string $language, string $fallback = 'en')
+    public function __construct(string $language, ?string $fallback = 'en')
     {
         $this->language = $language;
-        $this->fallback = $fallback;
+        $this->fallback = $fallback ?? false;
```
pkly: Why are you still using false here instead of `$this->fallback = $fallback ?? $this->fallback;` ?
There's always going to be an english string available so I think it should be somewhat forced

Also change the construct parameter to `?string $fallback = null`

I'd also suggest re-adding the `setLanguage` function, and simply calling it from the constructor.
That way it'd be possible to change the language during runtime (not very useful, but still)

abbadon1334: for the fallback, i don't commit the last change, it will be in the next push (today).
for the SetLanguage, i think there is no need of changing the language why allow it?

abbadon1334: i'm working on unit test now

abbadon1334: for the fallback i move the check for fallback===language in loading file from folder to avoid loading the same file twice, which is better

pkly:
@romaninsh are you planning on waiting with releasing 2.0 for this PR to get sorted out? 
It would be extremely nice to see it done, as it just seems to need few final tune-ups

pkly:
@abbadon1334 so in the end what are you having trouble with? using translations without an app or a parent chain?

abbadon1334:
> @abbadon1334 so in the end what are you having trouble with? using translations without an app or a parent chain?

now to use translatable you need to have AppScopeTrait or respect this minimal structure :
https://github.com/abbadon1334/core/blob/ceb466f09eccfd60626dd6ed91ace4a6057cb67d/tests/TranslatableTraitAppTest.php#L418-L435

Out of this you need work around to accomplish it. I think we need to speak about this.

For sure before adding i need to raise the level of Tests, and gain 100% like the other Traits

pkly:
> now to use translatable you need to have AppScopeTrait or respect this minimal structure 
> Out of this you need work around to accomplish it. I think we need to speak about this.

hm, well this seems kinda mediocre

I'll think about it on monday and write a mockup of how it could be implemented differently
I think there's something else that can be done, although it could require yet another trait (kinda like a TranslatorContainerTrait) or something.
That does seem like a bad way of doing it though.

Maybe add a function (inside the TranslateableTrait) that allows you to simply set the translator (a private member?)
That way you could assign the translator for an item, maybe also check if class is a container and apply the same translator object to the elements that this object contains?

romaninsh:
@pky this or can be merged as soon as it’s ready but before 2.0.0

romaninsh:
I’m ok to discuss this pky can you make it next Thursday?

Other thing I want to look at is string insertion. I am not a big fan of sprints and this can be same format as expr()

src/TranslatableTrait.php
romaninsh: ok, can we use `$this->__('we add {thing} into {container}', ['thing'=>'milk', 'container'=>'bowl'])` ?

That should be compatible with http://userguide.icu-project.org/formatparse/messages.

romaninsh:
I"m very far behind on the conversation here, I hope my feedback would still be relevant. Here are few things I'd like to have:

 - [ ] Use of arguments compatible with http://userguide.icu-project.org/formatparse/messages, e.g. `_('Hello {name}', ['name'=>'John'])`
 - [ ] Ability for user to implement a different translator in their app
 - [ ] Only use basic substitution in OUR messages
 - [ ] use namespace of 'atk4' or 'atk4\ui', but must be consistent. do we have to always specify it or can it be pulled out from the namespace? i mean - can trait get namespace of the class where it's used?

I don't see reason for translator containers or using multiple translators (as long as we have namespaces), but you can prove me wrong next meeting.

src/TranslatableTrait.php
```diff
@@ -0,0 +1,123 @@
...
+     *
+     * @param string $context the context domain if exists
+     * @param string $string  the string to be translated
+     * @param int    $count   the counter used to evaluate the plural
+     *
+     * @return string
+     */
+    public function _d(string $context, string $string, int $count = 1): string
+    {
+        if (isset($this->app) && method_exists($this->app, '_')) {
```
romaninsh: if we try using `self` here will it give us trait or class ?

abbadon1334: always class who use the trait, trait as a type not exists

romaninsh: that's awesome, can you then extract the namespace and use it as a context instead of `atk` ?

abbadon1334:
After the discussion on Gitter, i refactor the whole implementation to be used with Symfony Translation, can be used in and out of atk.

Need dependency of package Symfony\Component\Translation.

The doubts of @romaninsh about double plurals and other extreme cases that not all framework implements was right.
I agree that is better to stick with something external and largerly used, don't reinvent the wheel and discover only later that our wheel was already broken and we just don't see it.

In Symfony Translation Contracts repo, there are an Interface and a Trait that help to implement the library, i used the trait like they show, i mean to replicate the functionality when the TranslatorInterface is not present.

https://github.com/symfony/translation-contracts

For testing i just take data providers from Symfony Translation Tests and use in our 6 specific cases

- When method _ is called from  : 
- - (AppScope) An Object which use : AppScopeTrait+ContainerTrait+TranslatableTrait
- - (AppScopeChild) An Object which use : AppScopeTrait+TranslatableTrait that was add to (AppScope)
- - (AnyTranslatable) An Object which use : TranslatableTrait

My implementation was simple as we spoke at the beginning of the implementation, to try to avoid BC from Symfony i removed all extra helper methods. We are now transparent with Symfony implementation, i think this is good even for the documentation.

abbadon1334:
i want to be sure that conditions in Translatable::_ are respected and pass to right object even if TESTS are correct and show that lines are tested.

Test maybe change, but for me PR is final for now.

src/TranslatableTrait.php
```diff
@@ -0,0 +1,121 @@
...
+     * @var bool
+     */
+    public $_translatableTrait = true;
+
+    /**
+     * @var TranslatorInterface
+     */
+    private $translator;
+
+    public function setTranslator(TranslatorInterface $translator)
```
abbadon1334: i forget comment on method

src/TranslatableTrait.php
```diff
@@ -0,0 +1,121 @@
...
+    {
+        $this->translator = $translator;
+    }
+
+    public function getTranslator() : ?TranslatorInterface
```
abbadon1334: i forget comment on method

src/TranslatableTrait.php
```diff
@@ -0,0 +1,121 @@
...
+     * Check this property to see if trait is present in the object.
+     *
+     * @var bool
+     */
+    public $_translatableTrait = true;
+
+    /**
+     * @var TranslatorInterface
+     */
+    private $translator;
```
abbadon1334: Still think about if this private can be raise to protected, user must use SetTranslator to set and after that is all in the hand of method _ that will check to which object delegate the translations.

src/TranslatableTrait.php
```diff
@@ -0,0 +1,121 @@
...
+ * If a class use this trait, string can be translated calling method translate.
+ */
+trait TranslatableTrait
+{
+    /**
+     * Check this property to see if trait is present in the object.
+     *
+     * @var bool
+     */
+    public $_translatableTrait = true;
```
abbadon1334: added only for consistency with other traits but not used here

src/TranslatableTrait.php
```diff
@@ -0,0 +1,121 @@
...
+        // case if use TranslatableTrait and translator is defined
+        if ($this->hasTranslator()) {
+            return $this->translator->trans($id, $parameters, $domain, $locale);
+        }
+
+        // FALLBACK : Replicate internal library functionality
+        // if not use TranslatableTrait and is not a Child of App which use TranslatableTrait
+        $translator = new class() implements TranslatorInterface {
+            use TranslatorTrait;
+        };
```
pkly: Kind of wasteful to create a object every time, maybe add a static member, or a private one and initialize that if it's null?

pkly:
@romaninsh I can discuss it any time during bussiness hours (as long as I'm not working on another project) and maybe a bit later as well, but I don't really know how to get in touch outside of github and the forum, and this doesn't seem like the best way to talk really.

I'm fine with either sprintf-like syntax or tag-like syntax, the latter could be a bit more easier to change down the line. Though if you're gonna use that, probably just add a static method somewhere in core so that it's called by the translator (only tag parsing, not translation itself).

I would much prefer use of atk\core namespaces for translation, since that could add support for translation outside of using atk\ui.

There's no real need to use multiple translators, at least I don't think so, but I mentioned it as a possibility since we're running objects, so each one could have a different instance of a translator, and a different class all together.

Can't say I'm a huge fan of adding dependencies for everything for something that could be implemented in the library, but I won't really object, since it seems that this is just a few files.

@abbadon1334 
Also what's with this syntax?
```php
['There are 10 apples', '{0} There are no apples|one: There is one apple|more: There are %count% apples', 10],
```
I took a peek at the symphony code and I still don't quite get it.
I understand that the first text is the "real text" (?), not sure about the second string, and the number just means how many of an object there are(?)

abbadon1334:
> @abbadon1334
> Also what's with this syntax?
> ```
['There are 10 apples', '{0} There are no apples|one: There is one apple|more: There are %count% apples', 10],
> ```
> 
> I took a peek at the symphony code and I still don't quite get it.
> I understand that the first text is the "real text" (?), not sure about the second string, and the number just means how many of an object there are(?)

$expected => 'There are 10 apples'
$id => '{0} There are no apples|one: There is one apple|more: There are %count% apples'
$number => 10

which are used as dataprovider in this test : 
https://github.com/abbadon1334/core/blob/9eed1b48f6460ac2b3165eac65c5b952ee4cdbe2/tests/TranslatableTraitTest.php#L204

$app->_($id, ['%count%' => $number]); // if no translations, return $excepted

below equivalent definition : 
```php
// Symfony ICU-Message
            ['There are 10 apples', '{0} There are no apples|{1} There is one apple|]1,Inf] There are %count% apples', 10],
// Symfony STD
            ['There are 10 apples', '{0} There are no apples|one: There is one apple|more: There are %count% apples', 10],
```

pkly:
So one would use it as
```php
$text = '{0} There are no apples|one: There is one apple|more: There are %count% apples';
$res = $app->_($text, ['%count%' => 0]);
// $res: 'There are no apples'
$res = $app->_($text, ['%count%' => 1]);
// $res: 'There is one apple'
$res = $app->_($text, ['%count%' => 15]);
// $res: 'There are 15 apples'
```

Kinda like that?
Can't really say I like the syntax for different counts.

abbadon1334:
> So one would use it as
> 
> ```
> $text = '{0} There are no apples|one: There is one apple|more: There are %count% apples';
> $res = $app->_($text, ['%count%' => 0]);
> // $res: 'There are no apples'
> $res = $app->_($text, ['%count%' => 1]);
> // $res: 'There is one apple'
> $res = $app->_($text, ['%count%' => 15]);
> // $res: 'There are 15 apples'
> ```
> 
> Kinda like that?
> Can't really say I like the syntax for different counts.

if translator is defined and a translation entry is found for `'{0} There are no apples|one: There is one apple|more: There are %count% apples'` will return the correct translation, if translator is not defined it will process $text and return correct syntax.

pkly:
Right. I understand it now, but it's still a bit hard to get by simply looking at test cases, overall the implementation seems good but there's still the 'tag' question - does this support tag replacement for translations? I haven't seen the code that'd allow that, but if it's there then other than some docs with typical use cases it's fine.

abbadon1334:
> Right. I understand it now, but it's still a bit hard to get by simply looking at test cases, overall the implementation seems good but there's still the 'tag' question - does this support tag replacement for translations? I haven't seen the code that'd allow that, but if it's there then other than some docs with typical use cases it's fine.
> 
> Also you didn't answer the question about tags. Is that also supported here?
> (`'text {here} maybe', ['here' => 'me'] = 'text me maybe'`)

yes, and if there is a translation for that specific combination it will translate, if not it will only substitute.

i think more test is needed to show better usage and obviously some examples

src/TranslatableTrait.php
romaninsh: Wait,

TranslatableTrait is applied on ANY class that wants to use translations.

it adds method `_` which routes translations requests to APP or returns string as-is in case of a fallback. 

Why do we need translator defined here? Would that mean that pretty much all models and fields and whatnot will carry reference to translator?

src/TranslatableTrait.php
```diff
@@ -15,109 +18,104 @@ trait TranslatableTrait
...
-        $translated_args = [];
-        $translated_args[] = $this->_($string);
-        $this->_processTranslatedArray($translated_args, $args);
+        // case if is child Object which use AppScopeTrait + TranslatableTrait
+        // that was already add to an Object which use AppScopeTrait + ContainerTrait + TranslatableTrait
+        if ($this->hasTranslatorInAppScope()) {
+            return $this->app->_($id, $parameters, $domain, $locale);
+        }
+
+        // case if use TranslatableTrait and translator is defined
```
romaninsh: I don't think we should even implement this. It's overkill. 

Most of ATK UI will create 'app' if missing.

src/TranslatableTrait.php
romaninsh: Yeah, I think it's a misuse of "trait", in this case - just use of regular class would suffice, although I don't want ATK to handle the case when $app is undefined and should simply `return $id`;

src/TranslatableTrait.php
```diff
@@ -15,109 +18,104 @@ trait TranslatableTrait
...
-                    } else {
-                        $translated_args[] = $this->_d(...$sub_args);
-                    }
-                    break;
-                case 3:
-                    $translated_args[] = $this->_d(...$sub_args);
-                    break;
-            }
-        }
+        return isset($this->app) && $this->app !== $this && method_exists($this->app, '_');
```
romaninsh: I think this must be consistent with this:
`            if (!isset($this->app) || !isset($this->app->logger) || !$this->app->logger instanceof \Psr\Log\LoggerInterface) {
`
https://github.com/atk4/core/blob/develop/src/DebugTrait.php#L53

src/TranslatableTrait.php
romaninsh: I was thinking about if we want to have a fallback class here, but I still believe it's bad, because some users may end up relying on it and it would re-read language files on every invocation, which is an INSANE performance impact.

Either we use $app->translator or rely on service locator pr #99 inside App class. Either way - trait should assume that translator would be pre-initialized with the language files loaded.

romaninsh:
@abbadon1334 OK i see what you did there 😄 . you are using TranslatableTrait for both class-trait and a translator. Sorry you had to move things so much, please split this up:

 - trait we apply on most classes implementing '_' function relying on $app->translator
 - class that implements argument insertion for 'call {person} maybe', and being compatible with https://github.com/symfony/contracts/blob/master/Translation/TranslatorInterface.php (i don't particularly like extra dependency since it's not a PSR standard, just consistency is fine, but i don't feel strongly about it)
 - Test for the trait

As a test - we can try integrating this: "https://github.com/symfony/translation/blob/master/Translator.php#L195" into a mock app, then trying to translate.

Few last bits:

 - App must set Translator EXPLICITLY for now (otherwise _ will return $id)
 - This will change with #99, where we might change the way services are loaded

src/TranslatableTrait.php
abbadon1334: 1. translator can be removed, i understand your point and your statement the other day about not using property in traits, it became a dynamic property.
2. The idea was to use TranslatableTrait : 
- App which use TranslatableTrait, direclty calling method $this->_ 
- Objects add via add which use TranslatableTrait => calls $this->_ and delegate to $this->app->_
- Object that implements only TranslatableTrait => calling $this->_ 

if $translator is not an instance of TranslatorInterface, it will fallback to Symfony implementation that don't have any translation to be loaded, it just process the $id with $parameters and return a correct string with excepted substituition as intended in Symfony.

about this code : 
```php
$translator = new class() implements TranslatorInterface {
        use TranslatorTrait;
};
```
Initally i defined as a separate object and at that place there was something like : 
```php
if(!$this->translator) {
        $this->translator = new FallbackTranslator();
}
```
thinking about wrongly bypass the second condition. I think that can be better, because if we set a translator with SetTranslator, condition will stop before a check for is own proprerty $translator.
if no AppScope

abbadon1334:
> @abbadon1334 OK i see what you did there smile . you are using TranslatableTrait for both class-trait and a translator. Sorry you had to move things so much, please split this up:
> 
>     * trait we apply on most classes implementing '_' function relying on $app->translator

you said that you want to use $this->app->translator which means that translator can be done only using AppScopeTrait?

>     * class that implements argument insertion for 'call {person} maybe', and being compatible with https://github.com/symfony/contracts/blob/master/Translation/TranslatorInterface.php (i don't particularly like extra dependency since it's not a PSR standard, just consistency is fine, but i don't feel strongly about it)
i agree too, but i see that there are two wide used libs : Symfony or i18next.com, which are the only one that permits usage of translations without static reference with complete cover of cases like the ones we speak in chat

>     * Test for the trait
> 
> 
> As a test - we can try integrating this: "https://github.com/symfony/translation/blob/master/Translator.php#L195" into a mock app, then trying to translate.

you point me to a specific line you mean the method alone or the whole Symfony Translator? just to be sure

> Few last bits:
> 
>     * App must set Translator EXPLICITLY for now (otherwise _ will return $id)

that "return $id" kills me :) , because returning $id is easy, the problem is to replicate all the implementation of translator library and in the future remaining complaint without BC.

>     * This will change with #99, where we might change the way services are loaded
Probably translations can be loaded even with ConfigTrait, divided in files or folders. I'm probably too defensive, but adding things must be simple and strictly defined, if not i think is better to delegate it, like in this case, SetTranslator => i translate, No Translator => i return processed string from a sort of NullTranslatorInterface that replicate everything and return an excepted result.

Let's speak again tomorrow :D

abbadon1334:
>     * class that implements argument insertion for 'call {person} maybe', and being compatible with https://github.com/symfony/contracts/blob/master/Translation/TranslatorInterface.php (i don't particularly like extra dependency since it's not a PSR standard, just consistency is fine, but i don't feel strongly about it)

be consistent means implementing this, which i have done => https://github.com/symfony/contracts/blob/master/Translation/TranslatorTrait.php

@romaninsh you are right that if implemented like this, we are strictly coupling with it, but if we do it on our own, if there is a change we have to figure out what is wrong with our code.
IMHO if there is a good wheel that probably is less prone to be broken in the future, why not use it directly we are not coupling to the whole library only with contract and trait that make it working.

src/TranslatableTrait.php
pkly: Well, the fallback class should not ever translate anything, only parse the incoming string, I believe.
As for this specific line of code I'd simply suggest something like
```php
// where self::$_translationFallback = static private member defined to null

if (self::$_translationFallback === null) {
    self::$_translationFallback = new class() implements TranslatorInterface { 
        use TranslatorTrait;
    };
}

// use self::$_translationFallback
```

But that's only assuming that the TranslatorInterface/TranslatorTrait is required to parse the text, not actually translate it, or create this object somehow that it doesn't actually try to load language files.

abbadon1334: >But that's only assuming that the TranslatorInterface/TranslatorTrait is required to parse the text, not actually translate it, or create this object somehow that it doesn't actually try to load language files.

As you said, it only parse the arguments to return an excepted results.

My idea is that in this way we don't have to delegate to $this->app in case of external use, i think that $_translatorFallback will be called only if the Class not use AppScope or $translator is not set, just to have a fallback for usage of syntax excepted by Symfony\Components\Translation, that's it.

The idea was to use on every Object the same Trait and always use the method _ in place of $this->app, because in this case we have to move everything in AppScope or make 2 trait TranslatableTrait and TranslatorTrait.

If you make it today at 18:00 London, we have a meeting.
I don't know in which timezone you are, but you can come and partecipate, at this link : https://meet.google.com/dfb-vquj-kkk

pkly: Well yes, I understand why you did it and I prefer having the `_` function available in the classes themselves instead of calling `$this->app` constantly, that does seem like the better way to do it, since you can move any error checking to `_` as well.

I should be able to make it, I'm in +2 UT so it's not really an issue.

abbadon1334:
i started making some changes based on last discussions.
Sorry, but i was really busy in the last days and until monday i can't go back on this.
I pushed, just to get feedback if the actual direction of code is correct.

src/TranslatorSymfonyTrait.php
```diff
@@ -0,0 +1,57 @@
...
+
+    /**
+     * @param TranslatorInterface $translator
+     *
+     * @return TranslatorInterface
+     * @throws Exception
+     */
+    public function setTranslator($translator)
+    {
+        if ($translator instanceof TranslatorInterface) {
```
pkly: So Symfony/contracts will be a hard-dependency then?

abbadon1334: Has you see : src/TranslatorSymfonyTrait.php
is the Trait helper to use Symfony, if you want to use it you are free to use it.

every "bridged" translation class must pass from here :
```php
public function _($message, ?array $parameters = NULL, ?string $domain = NULL, ?string $locale = NULL): string
```

pkly: Well yes, I understand that, I think I understand it a bit better now that I looked a bit at the tests.
So basically you're creating an app and adding the translator TranslatorSymfonyTrait to it, otherwise (and by default) using the TranslateableTrait and it's minimal handling, correct?
But does that support the syntax that Symfony supports? This to be exact:
```php
'{0} There are no apples|]0,1[ There is almost one apple|{1} There is one apple|[1,Inf] There is more than one apple'
```
Your code simply explodes the text on `|` and assumes that each item is the next one in the counter, yes?

abbadon1334: take a look here : 
[https://github.com/abbadon1334/core/blob/wip-translation/tests/TranslatableTraitTest.php]
and here : 
[https://github.com/abbadon1334/core/blob/wip-translation/tests/TranslatorSymfonyTraitTest.php]

TranslatorSymfonyTrait => is only a helper to implement method `_` in App, setTranslator and  nothing more, to be used with a standard Symfony\Translation object

I'm working on rewrite the php library of I18next.com, actual implementation in php really sucks and doesn't support all the features, is used in many js project but support really easy even nested plural and substitution, the most interesting features is that accepts interpolation of token with array objects which can be interesting in atk4/data implementation

pkly: Ah, right.

Well from what I've seen their listed implementation is really basic, so... It actually looks a lot like my translator, which is fine for things that I'm using it for.

Still, you're just gonna use Symphony for translations for atk, right?

abbadon1334: From the last discussion we want to use an internal translation system but really basic, just to cover the few needs of actual atk4 translations, and leave any other complex implementation out of atk4.

i think this is the cases which at now will be covered, interpolation and simple plurals: 
https://github.com/abbadon1334/core/blob/e0c538c6179444e478cf711ae4baa1aa84147f99/tests/TranslatableTraitTest.php#L40-L53

PR #94, opened 2019-06-17
Feature/multi container trait

romaninsh:
That's the implementation of MultiContainer trait..


src/MultiContainerTrait.php
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Use this method trait like this:
+     *
+     * function addField($name, $definition) {
+     *     $field = $this->factory($definition, [], '\atk4\data\Field');
+     *
+     *     return $this->_addIntoCollection($name, $field, 'fields');
+     * }
+     *
+     * @param $name string Name that can be used to reference object
```
DarkSide666: order should be @param, then type and then name of argument

src/MultiContainerTrait.php
```diff
@@ -0,0 +1,108 @@
...
+        if (!$name) {
+            throw new Exception([
+                'Object must be given a name when adding into this',
+                'child'=>$object,
+                'parent'=>$this,
+                'collection'=>$collection
+            ]);
+        }
+
+        if (isset($this->{$collection}[$name])) {
```
DarkSide666: I think `array_key_exists` fits better here.

src/MultiContainerTrait.php
```diff
@@ -0,0 +1,108 @@
...
+     * Use this method trait like this:
+     *
+     * function addField($name, $definition) {
+     *     $field = $this->factory($definition, [], '\atk4\data\Field');
+     *
+     *     return $this->_addIntoCollection($name, $field, 'fields');
+     * }
+     *
+     * @param $name string Name that can be used to reference object
+     * @param $object mixed New element to add
```
DarkSide666: $object is not mixed type, it's `object` type

src/MultiContainerTrait.php
```diff
@@ -0,0 +1,108 @@
...
+        $this->{$collection}[$name] = $object;
+
+        // Carry on reference to application if we have appScopeTraits set
+        if (isset($this->_appScopeTrait) && isset($object->_appScopeTrait)) {
+            $object->app = $this->app;
+        }
+
+        if (isset($object->_trackableTrait)) {
+            $object->short_name = $name;
+            $object->name = $this->
```
DarkSide666: What's this???

src/MultiContainerTrait.php
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+
+        return $object;
+    }
+
+    function _removeFromCollection(string $name, string $collection) {
```
DarkSide666: comments

src/MultiContainerTrait.php
```diff
@@ -0,0 +1,108 @@
...
+
+        unset($this->{$collection}[$name]);
+    }
+
+    function _hasInCollection(string $name, string $collection) {
```
DarkSide666: comments

src/MultiContainerTrait.php
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * @param string $name
+     * @param string $collection
+     * @return object
+     * @throws Exception
+     */
+    function _getFomCollection(string $name, string $collection)
+    {
+        if (!isset($this->{$collection}[$name])) {
```
DarkSide666: Why not reuse `$this->_hasInCollection` method here? I know it can be very very tiny performance hit, but on other hand it will reuse code, so I think that would be good.
Like
```
if (! ($obj = $this->_hasInCollection($name, $collection)) ) {
    throw new Exception(...);
}
return $obj;

File: src/Exception.php [hunk 8]
```diff
@@ -149,7 +149,7 @@ public function getColorfulText()
             $output .= "\n\033[0mCaused by Previous Exception:\n";
             $output .= "\033[1;31m".get_class($p).': '.$p->getMessage()."\033[0;31m".
                 ($p->getCode() ? ' [code: '.$p->getCode().']' : '');
-            if ($p instanceof \atk4\core\Exception) {
+            if ($p instanceof self) {
```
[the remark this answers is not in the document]
romaninsh: i have checked with documentation and i think "self" is a correct way here. Basically it must use __CLASS__ or atk4/core/exception. If $p decends from it it's ok. 

If we use "static" than it would be the get_class($this), which could mean it's atk4\data\Exception and $p instanceof  will fail.

src/Exception.php
```diff
@@ -347,7 +347,7 @@ public function getHTML()
             $output .= '<div class="ui top attached segment">';
             $output .= '<div class="ui top attached label">Caused by Previous Exception:</div>';
 
-            if ($p instanceof \atk4\core\Exception) {
+            if ($p instanceof self) {
```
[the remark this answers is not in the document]
romaninsh: same as above.

src/MultiContainerTrait.php
```diff
@@ -0,0 +1,108 @@
...
+     * function addField($name, $definition) {
+     *     $field = $this->factory($definition, [], '\atk4\data\Field');
+     *
+     *     return $this->_addIntoCollection($name, $field, 'fields');
+     * }
+     *
+     * @param $name string Name that can be used to reference object
+     * @param $object mixed New element to add
+     * @param $collection string String corresponding to the name of the property
+     * @return $obect
```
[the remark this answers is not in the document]
romaninsh: fixed.

src/MultiContainerTrait.php
romaninsh: fixed.

src/MultiContainerTrait.php
romaninsh: fixed.

src/Exception.php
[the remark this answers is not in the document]
abbadon1334: i confirm behaviour of self over static. +1 for self

src/MultiContainerTrait.php
```diff
@@ -0,0 +1,182 @@
...
+     * @param object $object     New element to add
+     * @param string $collection string String corresponding to the name of the property
+     *
+     * @throws Exception
+     *
+     * @return object|mixed $obect
+     */
+    public function _addIntoCollection(string $name, object $object, string $collection)
+    {
+        if (!$collection || !isset($this->{$collection}) || !is_array($this->{$collection})) {
```
abbadon1334: as argument is defined as string, can be better check for empty(!$collection)?

src/MultiContainerTrait.php
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * @param string $name
+     * @param string $collection
+     *
+     * @throws Exception
+     *
+     * @return object
+     */
+    public function _getFomCollection(string $name, string $collection)
```
abbadon1334: typo : getFromCollection

src/Exception.php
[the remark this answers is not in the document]
DarkSide666: ok

src/MultiContainerTrait.php
romaninsh: And what about objects

abbadon1334: object LGTM => https://wiki.php.net/rfc/object-typehint

for empty @DarkSide666 you are right but i never used this way, i checked and is OK : 
http://sandbox.onlinephpfunctions.com/code/e9f41cac01daf4c17a583baa11f07e5503deebed

src/MultiContainerTrait.php
romaninsh: Well spotted!

Issue #87, opened 2019-02-25
Add docs for ConfigTrait


DarkSide666:
ConfigTrait was implemented here https://github.com/atk4/core/pull/86, but there are no docs for it.

romaninsh:
probably there should also be an adequate example in atk4/ui ?

PR #98, opened 2019-06-20
Fix config trait file bad format

abbadon1334:
i make a draft for documentation of ConfigTrait ( ref : #87 )
in my project as i show you last week, i extensively use ConfigTrait with DefinitionTrait ( You will see it in a next PR ).
If you try to load an empty file, you will get an error and not an exception on : 

https://github.com/atk4/core/blob/develop/src/ConfigTrait.php#L83

because $tempConfig is not an array, to avoid that i add a check against type array and raise exception if not.


docs/config.rst
```diff
@@ -0,0 +1,83 @@
...
+    $object = new Object();
+    $object->readConfig('config.php','php');
+
+After loading, configuration elements can be retrieved in this way :
+
+    $object->getConfig('element_key','if not defined use this as default');
+
+if you need an element that is declared inside an array you can use a special syntax :
+
+    $object->getConfig('level1_array/level2_array/element_key','if not defined use this as default');
```
romaninsh: i think we should also provide a good guideline on how 3rd party add-ons should use their class as a top level element key.

abbadon1334: Translator implementation for sure will need it and can be created as an addon.

At this moment the TranslatableTrait return translations already written in codebase, using an addon we can extract all this "tokens" and define it as an addon.

I think we probably need a composer script or a console command that create the folder in the project for you when add the addon.

next thursday we can speak about how can be implemented addons when need extra config, like we have done with actions, the results was really good.

docs/config.rst
```diff
@@ -0,0 +1,83 @@
...
+
+    $config=[
+        'var A' => new UserClass(),
+        'var B' => 2,
+        'var C' => [
+            '2nd-level' => 'var D'
+        ]
+    ];
+
+php-line
```
DarkSide666: php-inline

mvorisek:
@abbadon1334 What about unifying support for PHP config to support only return?

PR #99, opened 2019-06-20
ServiceLocatorTrait

abbadon1334:
Take it as a draft of the functionality discussed in the last meeting about definition of Object directly in config.

I add Documentation and Test and improve my actual version that i already use in my project.

is ispired by https://pimple.symfony.com/
i called it Definition and Definer because we already have our naming for DI and Factory which is used internally in atk4 and i don't want to disorient.

I think this can be seen as good helper for loading external dependency in full atk4 projects or at least is what i do in mine.

if this thunderstorm don't drop my office, we can speak about it later in the meeting.


docs/definition.rst
```diff
@@ -0,0 +1,185 @@
+=====================
+Definer/Definition Trait
+=====================
+
+.. php:trait:: DefinerTrait
```
romaninsh: from the documentation here it's not obvious what the benefit is.

My understanding is vague. 

I think it allows you to use a 3rd party app class without extending it or writing any code. Injecting it through config files, so you have to define this pattern.

It's not usable for ATK itself and from the initial doc it looks like it does same thing as factory but more complex and unorthodox. Try looking at it as someone who is not familiar to Pimple 🤯

abbadon1334: I just adapted something that i already use and is really good to have all configurations or classes in a folder distributed in multiple files, that can be called with only one line of code only when needed, if first argument is FQCN the editor probably can do something for intellisense. Obviously for a small project can be only something in plus to be defined, but for medium to big project i think having configurations in a single place is a must with this can be done with lazy object instantiation. i'm sorry if in docs i was not so good to pass the concept and trust me is not so unorthodox PHP-DI and Pimple are used everywhere in php community.

abbadon1334:
I analyzed it better and is clearly a Service Locator Design Pattern, which if misused can lead to anti-pattern.

I will look further and use it more deeply in my applications, to provide examples and a better documentation.

I added a label "in stand by"

docs/definition.rst
romaninsh: The implementation itself is good, but I still struggle to see why not simply do:

```
$app->logger = include 'config/logger.php';
```

It's not like we'll want multiple loggers in our app. Is there any other example? I'd like to hear from others that:
 - yes, we can't live without it!
 - yes, it must be in ATK core, can't be elsewhere!

abbadon1334: @romaninsh I'm not sure too, just figure out if it can be useful, i see more and more glitch trying to be compliant with PSR-11 ContainerInterface, just too many repetition in code.

For me can be closed.

romaninsh: @abbadon1334 lets still discuss it. There is some possibility it can help us initialize $app->logger and $app->translator and other things like that.

It's not a very good practice to rely in atk/core on non-existing property of non-existing object.

abbadon1334:
I remain on the same vision, i mean there are many other ways to add something out of atk4 and setup it with initializer when add.

I think we already have half of this but "with other names" / "with other ways to implement", adding this is like adding a new way of adding something to a container.

I still leave here as it is "on hold", to let us rethink about having : 
 - ->add(new Something()) that provide an extra feature that must be called to be used.

In short if we add this we start asking too much, like : "this service is enabled? use it : use another", in place of tell "do this! and wait for a response if needed", all condition/execution must happen internally.

And if we add a new concept like Service, which can be Email ( outbox is there as a draft), Logging, Translations, SMS service, external SaaS, all needs an external configuration and an object/service to be used.

Every service needs : 
 - Configuration
 - Check for Models
on init if a flag Service::$dev is true, needs to check for creation/sync of persistence tables

Every service provide : 
 - Views
this is not a problem, cause they will be called in the app when needed
 - Models
 - Hooks
interesting to have a way of defining hooks that for sure will couple the system, but can be activated to automate integration of some Service ( e.g. audit is hardly coupled on persistence save )

Login/Audit/Language if abstracted to what they do/needs/provide are very similar, but are not normalized and defined as a concept to be followed, i think can be interesting to define a sort of Abstract Service as a starting point for adding services to atk4?

ATK4 is like LEGO, we already have all components to build higher level abstraction components, the problem many times is there are no instructions or concept to follow, this degree of freedom in my opinion make discussion feels too much technical or too much abstract and lead to misunderstanding and loosing of time thinking about how to start to contribute.

What do you think guys ? Take a look at the below code, because i think that all my PR can be reduced to this few lines : 

```
/**
 * I think the only need for this is to work
 * is to add a method App->addService(Service $service)
 * to bypass the limitation of method App->add that needs $app->layout to be set,
 * but now this can be done with CollectionTrait in App
 */
abstract class ServiceProvider {

    use \atk4\core\AppScopeTrait;
    use \atk4\core\DIContainerTrait;
    use \atk4\core\InitializerTrait;

    /** @var \ServiceConfigurator */
    private $config;

    /** @var string */
    protected $config_file;

    /** @var string */
    protected $config_type;

    /** @var !!redefined type in every extended classes  */
    protected $service;

    public function __construct(array $defaults = [])
    {
        $this->config = new ServiceConfigurator();

        $configs = $defaults['config'] ?? [];
        foreach($configs as $key => $value)
        {
            $this->config->setConfig($key,$value);
        }

        unset($defaults['config']);

        $this->setDefaults($defaults);
    }

    protected function loadConfig()
    {
        if(null !== $this->config_file && null !== $this->config_type) {
            $this->config->loadConfig($this->config_file, $this->config_type);
        }
    }

    public function init()
    {
        $this->loadConfig();
    }

    public function getService()
    {
        return clone $this->service;
    }
}

class ServiceConfigurator {
    use \atk4\core\ConfigTrait;
}
```

Issue #108, opened 2019-07-18
Port "addSolution" to atk core Exceptions and UI 


romaninsh:
be considerate of #93 ..

https://github.com/atk4/atk4/blob/4.4/lib/BaseException.php#L137

abbadon1334:
I think this can be really interesting

pkly:
Somewhat related: I was thinking about allowing a few other things to be added/changed in the exceptions.
Name being one of them, since now it will always say "Agile Toolkit Exception".
It'd be rather easy to change that to pull the name from a variable (protected) and that'd allow you to create new exceptions that inherit from atk4/core and override that name for better user experience.

It'd also be nice to add some sort of data that can be stored with the exception - currently you can only add the params, which is not ideal - it'd be better if you'd allow data that doesn't need keys, simply the value and then display it to the end-user.

pkly:
Fixed by #109 so this should be closed.
@romaninsh

PR #109, opened 2019-09-27
Exception->addSolution and some other improvement

abbadon1334:
- add method addSolution to Exception.
- separate Logic from Rendering
- add HTML Exception Renderer
- add HTMLText Exception Renderer
- add Console Exception Renderer
- add JSON Exception Renderer (new)
- add Tests for addSolution
-  #108 Custom Exception name can be defined using `$custom_exception_name` (@pkly)

We can do more on Solution, like in https://github.com/atk4/atk4/blob/4.4/lib/BaseException.php#L137 but we already have params and solution, adding a button to see a query in detail can be accomplished using params, for now.

How the solution is rendered in different formats can be seen in the screenshots below 
HTML
image
HTMLText
image
Console
image
JSON
image


src/Exception.php
```diff
@@ -406,10 +168,58 @@ public function getParams()
...
+     *
+     * @return string
+     */
+    public function getCustomExceptionName(): string
+    {
+        if ($this->custom_exception_name) {
+            return $this->custom_exception_name;
+        }
+
+        return get_class($this);
```
pkly: Since atk4/core is targetting 7.2+ why not just 
`return $this->custom_exception_name ?? get_class($this);`

abbadon1334: right!

abbadon1334:
@pkly > Still no custom data is allowed to be passed to the exception

With `addSolution` you can add a string to suggest a solution, with `addMoreInfo` you can add a key and a value, if value is not an object, will be normalized to json string otherwise will be replaced with the object FQCN.

Display any type of data can be tricky sometimes, one way to accomplish this is to leave the developer to call `->addMoreInfo` with a value already normalized or create a method `->addMoreData($value)` that transform $data to json string.

What you think?

pkly:
Right, I guess I must've missed it a bit. `addSolution` is rather nice, and should be fine for the time being. As for `addMoreInfo` - my gripe with it is that it's keyed, so you have to think of some names that might not really make much sense to the end user who's going to see this information.

I suppose this solution can work for the time being, but still. As for the json-ification - why not do a check like this:

```php
if (!is_string($value)) {
    if (is_object($value) && method_exists($value, '__toString'))
        $value = $value->__toString();
    else if ($value instanceof \JsonSerializable)
        $value = json_encode($value);
    else
        $value = (string)$value; // unsure about this tho, maybe something else?
}
```

I'm not quite sure how to best solve the other issue, but for the time being it's not a _huge_ problem, so it's fine to ignore it.

abbadon1334:
this is done internally by PHP `if (is_object($value) && method_exists($value, '__toString'))` when you force the typecast `(string) $value`

probably can be simplified like this :
```php
$value = $value instanceof \JsonSerializable ? json_encode($value) : (string) $value;
```
The main problem on exception was that if they are managed incorrectly can disclose security issues, for sure, we will speak about this subject and localization on next meeting.

pkly:
I specifically listed the `__toString` check first _because_ it could be a security issue if you simply throw everything into json. Besides, the string values should be more useful in a lot of cases than simply what json would need, because you could write an actual message inside the class, instead of simply printing it's values.

pkly:
> just try to do add solution...
> 
> Now to add a solution you need to :
> 
> ```
> throw (new Exception([
>     'error blah blah',
>     'table' => 'db_table',
>     'query' => $db->getQuery()
> ])->addSolution('this is the solution');
> ```
> 
> i don't like the format, why not defining a special word without BC all the actual code :
> 
> ```
> throw new Exception([
>     'error blah blah',
>     'table' => 'db_table',
>     'query' => $db->getQuery(),
>     'solution' => ['this is the solution']
> ]);
> ```
> 
> @romaninsh what you think?

Why not both? It's not like they'd interfere with each other.

romaninsh:
@pkly it's not BC breaking. We are not breaking any existing code, just adding new method(s).

PR #117, opened 2020-01-12
Configuration Elements cannot be replaced

bepixeld:
If multiple configuration files are used, a configuration item cannot be replaced.
e.g. Standard Configuration with User Configuration.
default: `$config['lang' => 'en'];`
user: `$config['lang' => 'de'];`
Calling `$obj->getConfig('lang')` returns an array `['en', 'de']`.
However, the string 'de' would be desired.
Is there a special reason why `array_merge_recursive` is used?
This can be prevented with `array_replace_recursive`.
Thanks in advance for an explanation.

### DIFF
--- a/src/ConfigTrait.php
+++ b/src/ConfigTrait.php
@@ -94,7 +94,7 @@ trait ConfigTrait
             $configs[] = $tempConfig;
         }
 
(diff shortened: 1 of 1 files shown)


DarkSide666:
fix #118

PR #119, opened 2020-02-03
Add injection of solutions in Exception constructor

abbadon1334:
add injection of solutions in constructor.
Now to throw an exception and add a solution you have to :
```php
$e = new Exception(['Exception with solution']);
throw $e->addSolution('One Solution');
// OR
throw (new Exception(['Exception with solution']))->addSolution('One Solution');
// OR
throw (new Exception(['Exception with solution']))
        ->addSolution('1st solution')
        ->addSolution('2nd solution')
;
```
With this PR it can be done this way :
```php
throw new Exception([
        'Exception with solution',
        'solutions' => 'Solution to exception'
]);
// OR
throw new Exception([
        'Exception with solution',
        'solutions' =>[
                '1st solution to exception'
                '2nd solution to exception'
        ]
]);
```
Not a big change but a much _atk syntax_


src/Exception.php
```diff
@@ -62,6 +62,15 @@ public function __construct(
             // message contain additional parameters
             $this->params = $message;
             $message = array_shift($this->params);
+            if ($this->params['solutions'] ?? null) {
```
mvorisek, line 65: use `isset()` here

src/Exception.php
```diff
@@ -62,6 +62,15 @@ public function __construct(
...
+                if (!is_array($this->params['solutions'])) {
+                    $this->params['solutions'] = [$this->params['solutions']];
+                }
+                foreach ($this->params['solutions'] as $solution) {
```
mvorisek, line 69: `foreach ((array)$this->params['solutions'] as $solution)` should be enought, no need for the first `if`.

PR #128, opened 2020-02-25
avoid code duplication

georgehristov:


src/HookTrait.php
```diff
@@ -58,29 +58,19 @@ public function onHook($spot, $fx = null, $args = [], $priority = 5)
...
-                    if (!method_exists($fx, $spot)) {
-                        throw new Exception([
-                            '$fx should be a valid callback',
-                            'fx' => $fx,
-                        ]);
-                    }
-                }
-                $fx = [$fx, $spot];
-            } else {
+                $valid = isset($fx->_dynamicMethodTrait) && $fx->hasMethod($spot) || method_exists($fx, $spot);
```
PhilippGrashoff: what do you think about addingbrackets to clarify logic?
`$valid = (isset($fx->_dynamicMethodTrait) && $fx->hasMethod($spot)) || method_exists($fx, $spot);`

georgehristov: 👍 everything that makes code speak for itself

georgehristov:
Tests are failing after last update on tests?

PR #129, opened 2020-03-08
Give a hint why getElement() could have failed

NotAProfessionalDeveloper:
Help the user in finding stupid mistakes they make

### DIFF
--- a/src/ContainerTrait.php
+++ b/src/ContainerTrait.php
@@ -246,8 +246,12 @@ trait ContainerTrait
     public function getElement($short_name)
     {
         if (!isset($this->elements[$short_name])) {
+            $msg = 'Child element not found';
+            if ($short_name == 'name') {
+                $msg .= " (Hint: 'name' is the default for a model's \$title_field property. Did you maybe forget to set it?)";
+            }
             throw new Exception([
-                'Child element not found',
+                $msg,
                 'parent'  => $this,
                 'element' => $short_name,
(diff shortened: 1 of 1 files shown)


NotAProfessionalDeveloper:
An error about getElement() not being able to retrieve the "name" field suggests that the user forgot to define the title field in the model. Since we know that forgetting to add a title field would end up in this (for the user, cryptic) message, the software could give a hint about what's usually wrong, and therefore lowering frustration.  Maybe there is a place in atk4/data to put this in. Never mind.

PhilippGrashoff:
We have a feature to exceptions which can suggest a possible solution, @abbadon1334 implemented it. This seems a very good use case for it, adding a possible solution with the message you proposed, maybe with the addition "If you use atk4/data, ...."

DarkSide666:
Yes this could be good to use `addSolution()` in this case, but anyway this should not go in core. Right place to set this would be somewhere in atk4/data.
Also if I remember correctly, then `id_field` is used automatically in case `title_field` is not found.
See in atk4/data/Model:
```
    /**
     * Return value of $model[$model->title_field]. If not set, returns id value.
     *
     * @return mixed
     */
    public function getTitle()
    {
        if (!$this->title_field) {
            return $this->id;
        }
        $f = $this->hasField($this->title_field);

        return $f ? $f->get() : $this->id;
    }
```

PR #133, opened 2020-03-13
Improve hooks and allow to delete/check specific priority

### LINKED ISSUE #131: Multiple hooks for one action can be added, but they have to be currently deleted completely
See https://github.com/atk4/core/blob/develop/src/HookTrait.php#L98

Specific hook should be possible to delete by name and priority.

mvorisek:
Fixes #131

### DIFF
--- a/src/HookTrait.php
+++ b/src/HookTrait.php
@@ -21,9 +21,9 @@ trait HookTrait
     /**
      * @deprecated use onHook instead
      */
-    public function addHook($spot, $fx, $args = null, $priority = null)
+    public function addHook($spot, $fx, array $args = null, int $priority = null)
     {
-        return $this->onHook(...func_get_args());
+        return $this->onHook($spot, $fx, $args, $priority ?? 0);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


mvorisek:
Fix composer/tests first, see https://github.com/atk4/core/issues/134

PhilippGrashoff:
Well, this solution is perfectly backwards compatible and easy, but also not perfectly save to use :) There could be more than one hook with the same priority set, and so one could accidently delete more hooks than intended. But this chance is definitely low when passing a custom priority to a hook. To me, this is good enough...

mvorisek:
I analysed this more deeply and returning an index under which the hook was saved seems to be the best solutions for later manipulation. This and only this concept prevents collisions.

src/HookTrait.php
```diff
@@ -18,32 +18,37 @@ trait HookTrait
...
+    private $_hookIndexCounter = 0;
+
     /**
      * @deprecated use onHook instead
      */
-    public function addHook($spot, $fx, $args = null, $priority = null)
+    public function addHook($spot, $fx, array $args = null, int $priority = null)
     {
-        return $this->onHook(...func_get_args());
+        throw new Exception('Method addHook is deprecated, use onHook instead.');
```
DarkSide666: This is to big backward compatibility break and should be done only on next major release.
For now I think addHook should stay as proxy method for compatibility.

mvorisek: @DarkSide666 Yes, it seems so, I will remove that exception.

What about introducing some static method that can be easily called in deprecated code to emit deprecated warning/php notice? Otherwise almost all people ignore deprecations until the deprecated feature is removed.

DarkSide666: Yes that could be useful I guess.

PR #142, opened 2020-03-15
Add default class name fallback for addToWithClassName method

mvorisek:
PR done - but analyse first if `static::class` or `self::class` or none (no PR) should be used.

If used from static code, then `static::class` is perfect.

But if used from some method like:
```
public function add($seed) {
    return static::addToWithClassName($this, $seed);
}
```
on extended class like `CRUD`, then `$crud->add(['ui' => 'something'])` will create `CRUD` instance instead of `View`.

There should be probably no default class.

### DIFF
--- a/src/StaticAddToTrait.php
+++ b/src/StaticAddToTrait.php
@@ -112,15 +112,15 @@ trait StaticAddToTrait
                 $seed = [$seed];
             }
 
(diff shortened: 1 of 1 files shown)


mvorisek:
There should be probably no default class.

PR #152, opened 2020-04-08
Shorten exception trace

mvorisek:
Before:
image

image

After:
image

image

Example code:
```
    try {
        throw new \atk4\ui\Exception('source');
    } catch (\atk4\ui\Exception $ex) {
        throw new \Exception('parent', 0, $ex);
    }
```

### DIFF
--- a/src/ExceptionRenderer/Console.php
+++ b/src/ExceptionRenderer/Console.php
@@ -86,10 +86,9 @@ TEXT;
 
         $in_atk = true;
         $escape_frame = false;
-        $tokens = [];
-        $trace = $this->exception instanceof Exception ? $this->exception->getMyTrace() : $this->exception->getTrace();
-        $trace_count = count($trace);
-        foreach ($trace as $index => $call) {
+        $short_trace = $this->getStackTrace(true);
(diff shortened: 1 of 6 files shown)


romaninsh:
pls look into https://codeclimate.com/github/atk4/core/pull/152 too, if anything can be done there, not a problem either way]!

abbadon1334:
> pls look into https://codeclimate.com/github/atk4/core/pull/152 too, if anything can be done there, not a problem either way]!

I see duplication can be avoided but i think that added complexity and the other one on conditional can be changed and defined as variables but i like is more workaround than a solution to avoid triggering codeclimate logic.

¯\_(ツ)_/¯

mvorisek:
Yes, I will fix the large if and leave the duplicated code. It is not from ;-) - the exception renderers should actually be merged into one class as currently there is no way to extend them (or it needs to be done for every of them).

mvorisek:
Even this was too huge for codeclimate - https://github.com/atk4/core/commit/7b8fa1b44ce05893ec22ad49e91b41930c1826fa ... Of course, it can be solved for ex. with foreach, but that will be not better that the current code.

PR #153, opened 2020-04-09
Remove custom alias/name for exception class

mvorisek:
Plus use "Critical Error" for non-atk exceptions to be consistent with them.

Before:
image

After:
image

### DIFF
--- a/src/Exception.php
+++ b/src/Exception.php
@@ -28,9 +28,6 @@ class Exception extends \Exception
     /** @var string */
     protected $custom_exception_title = 'Critical Error';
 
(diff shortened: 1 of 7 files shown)


romaninsh:
Added by @abbadon1334 , I don't think it's documented though. Not sure what was the intention.

mvorisek:
@romaninsh Having custom name in opposite to the real class name is very confusing. I left there custom_exception_title even it seems to be not used acrosss in atk projects.

romaninsh:
@mvorisek i don't agree or disagree - just want to give a chance to  @abbadon1334 and hear his reasoning ;)

PR #156, opened 2020-04-11
Move types to code if possible

### LINKED ISSUE #154: assertRegExp() is deprecated and will be removed in PHPUnit 10. Refactor your code to use assertMatchesRegularExpression() instead.
There are several warnings in current tests which needs to be fixed/refactored.

see https://github.com/atk4/core/runs/573845743#step:10:15

mvorisek:
And also fixes #154

Forcing `:self` return type instead of phpdoc `@return $this` is not desired as some IDEs the ignore the phpdoc and uses `self` instead of `static` type. This will change in PHP 8.0, which will support native return type of `static`.

### DIFF
--- a/docs/factory.rst
+++ b/docs/factory.rst
@@ -405,7 +405,7 @@ values. See my description below the example::
     class RedButton extends Button {
         protected $icon = 'book';
 
-        function init() {
+        function init(): void {
             parent::init();
 
             $this->icon = 'right arrow';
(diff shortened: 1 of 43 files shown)


mvorisek:
Rebased on the latest `develop`, PR done and ready to be merged.

romaninsh:
Some of the previous commits broke localisation exception tests. Could you see why is it failing?

PR #157, opened 2020-04-13
Modernize coding style

mvorisek:
Implement PSR12


mvorisek:
@romaninsh , @abbadon1334 Can we configure StyleCI to use CS from `.php_cs.dist` directly?

abbadon1334:
> @romaninsh , @abbadon1334 Can we configure StyleCI to use CS from `.php_cs.dist` directly?

No it has is own configuration file, we need to find a sync between StyleCI and PHPCS and set all repos to that.

PR #159, opened 2020-04-14
Unify test method header

mvorisek:
image


DarkSide666:
how does this help?

mvorisek:
@DarkSide666 Default parameter value is there like a typo, only cleanup.

DarkSide666:
Ah yes, right.
I guess I need a sleep - was mixing red and green color :)

PR #160, opened 2020-04-14
Fix "driver" to "driver type" change

mvorisek:
Retest/merge after https://github.com/atk4/data/pull/559


mvorisek:
> 
> 
> tests still fail on php 7.2, but not on php 7.3

Really? (reopened to trigger tests using the updated translations)

PR #165, opened 2020-04-15
Fix ResultPrinter class compatibility with phpunit < 6.0

mvorisek:
Why?
https://travis-ci.org/github/atk4/ui/jobs/675692696#L731


mvorisek:
@DarkSide666 Extremely important to support phpunit 5.x used almost everywhere and the latest phpunit 9.x. Only this way the CI can be kept green before all repos are fully migrated to the new phpunit. Once done, the support for the old versions can be dropped.

DarkSide666:
OK I see.

PR #168, opened 2020-04-21
Improve assertions counter

mvorisek:


mvorisek:
Do we want to suppress no tests at all?

PR #174, opened 2020-04-23
Append missing slash to locale directory

mvorisek:


src/Translator/Adapter/Generic.php
```diff
@@ -95,7 +95,7 @@ protected function loadDefinitionATK(string $locale): void
 
         if (class_exists('\atk4\data\Locale')) {
             $path = Locale::getPath();
-            $this->addDefinitionFromFile($path . $locale . '/atk.php', $locale, 'atk', 'php-inline');
+            $this->addDefinitionFromFile($path . '/' . $locale . '/atk.php', $locale, 'atk', 'php-inline');
```
georgehristov, line 98: I suggest replacing `/` with `DIRECTORY_SEPARATOR` here

mvorisek: I already fixed your feedback for this in https://github.com/atk4/ui/pull/1124 , but it should be not needed and as you can see here - a few chars later :) there is `/`.

PS: `/` does work on WIndows perfectly, if there are any issues is general, let me know

georgehristov: We can make it better than it was (replacing the existing `/`).
It is just to keep it being OS agnostic as far as possible

mvorisek: It is up too you to fix this, I will not propose these changes.

PR #175, opened 2020-04-23
Normalize Unit Testing workflow across repos

mvorisek:


mvorisek:
PR is done and more stricter syntax is fully compatible with current StyleCI settings.

@DarkSide666 can you review?

src/DebugTrait.php
```diff
@@ -131,8 +128,6 @@ public function debugTraceChange(string $trace = 'default'): void
      * System is unusable.
      *
      * @param string $message
-     *
-     * @return void
      */
     public function emergency($message, array $context = [])
```
DarkSide666, line 132: Why not moving type `string $message`  to method parameters instead of phpdoc comment?
In other places you do like that.
Same goes for all these similar methods below.

src/Exception.php
```diff
@@ -199,8 +199,6 @@ public function getCustomExceptionTitle(): string
     /**
      * Set Custom Translator adapter.
      *
-     * @param ITranslatorAdapter|null $adapter
-     *
      * @return $this
```
DarkSide666, line 202: no more need for `@return $this` comment I guess.

src/ExceptionRenderer/RendererAbstract.php
```diff
@@ -134,7 +134,7 @@ protected function getStackTrace(bool $shorten): array
 
         $both_atk = $this->exception instanceof Exception && $this->parent_exception instanceof Exception;
         $c = min(count($trace), count($parent_trace));
-        for ($i = 0; $i < $c; $i++) {
+        for ($i = 0; $i < $c; ++$i) {
```
DarkSide666, line 137: is `++$i` somehow better than `$i++` here?

src/TrackableTrait.php
```diff
@@ -38,7 +38,7 @@ trait TrackableTrait
      */
     public function getDesiredName(): string
     {
-        return preg_replace('/.*\\\\/', '', strtolower(get_class($this)));
+        return preg_replace('/.*\\\\/', '', strtolower(static::class));
```
DarkSide666, line 41: is `static::class` somehow better than `get_class($this)`?

src/DebugTrait.php
mvorisek: Because of  parent interface :)

src/ExceptionRenderer/RendererAbstract.php
mvorisek: I do not like it too, but for some reasons it is prefered by CS fixed by its rules - so if you agree, I can add one extra CS rule to change it to the original form.

src/TrackableTrait.php
mvorisek: Yes, it is instantly optimalized by PHP at compile time, changed by one of the default CS rules.

src/ExceptionRenderer/RendererAbstract.php
DarkSide666: Described here: https://stackoverflow.com/a/1756021/1466341
At least 10 years ago they claim that pre-increment was 10% faster than post-increment.

src/Exception.php
mvorisek: `self` means `self`, `$this` means "instance of `static`". I would love to change this to `: static`, but it is supported as of PHP 8.0 :D

Last week I did quite large research on this for maximum compatiility and we should not add `: self` to classes where refactoring of child classes need to be supported.

Many IDEs prefer the return type in the code and does not hint any methods/properties from the child classes even if the phpdoc is presented. Just to let you know and due this I withholded changes of object return types in data, especially for `Model` which need to support full refactorability for child classes.

DarkSide666: OK, I see.

src/ExceptionRenderer/RendererAbstract.php
DarkSide666: Tested: `++$i` is always faster by 10-25% than `$i++` so let's stick with `++$i` where possible.

PR #178, opened 2020-04-27
Normalize this repo from others

### LINKED ISSUE #176: Merge order should not change order of array seeds
This fails:
https://github.com/atk4/core/blob/47c62bd335b08dfae478d9ce7a253e3571a67195/tests/SeedTest.php#L184

if `assertSame` is used with the following diff:
```
4) atk4\core\tests\SeedTest::testMerge6
Failed asserting that two arrays are identical.
--- Expected
+++ Actual
@@ @@
 Array &0 (
+    4 => 'four'
     5 => 'five'
-    4 => 'four'
 )

/__w/core/core/tests/SeedTest.php:181
/__w/core/core/src/AtkPhpunit/TestCase.php:15
```

mvorisek:
- `single_line_throw` is very bad with Atk exceptions when parameters are added - there should be kept on multiple lines.
  - note: there were no multiline throws in this repo even before https://github.com/atk4/core/pull/175
- fix #176


tests/SeedTest.php
```diff
@@ -105,23 +105,23 @@ public function testMerge3()
         $oo = $this->mergeSeeds(['foo' => 1], $o);
 
         $this->assertSame($o, $oo);
-        $this->assertSame($oo->foo, 1);
+        $this->assertSame(1, $oo->foo);
```
mvorisek, line 108: Why? reference value should come always first

codecov.yml
```diff
@@ -6,11 +6,11 @@ ignore:
 coverage:
   precision: 2
   round: down
-  range: "95...100" # lower coverage expectations
+  range: "95...100" # coverage expectations
   status:
     project:
       default:
-        threshold: 0.5% # lower coverage diff threshold, but maybe is not working
+        threshold: 1% # coverage diff threshold
```
mvorisek, line 13: remove both comments if it works, they are clear...

src/FactoryTrait.php
```diff
@@ -133,11 +136,17 @@ public function factory($seed, $defaults = [], string $prefix = null): object
...
 
+        $object = array_shift($arguments); // first numeric key argument is object
+        if (!$object) {
+            throw new Exception([
+                'factory() could not find object',
+                'seed' => $seed,
+            ]);
+        }
+
         if (is_string($object)) {
```
mvorisek, line 150: @DarkSide666 if we put `!is_object($object)` here, would it work? We can then remove the if condition above, it is bad, as it emits about the same meaning with completely different message.

DarkSide666: will `normalizeClassName(null, $prefix)` not throw exception then?
it will because `normalizeClassName()` only accepts 1st parameter as string type and nothing more.

src/FactoryTrait.php
```diff
@@ -142,7 +142,7 @@ public function factory($seed, $defaults = [], string $prefix = null): object
         $object = array_shift($arguments); // first numeric key argument is object
         if (!$object) {
             throw new Exception([
-                'factory() could not find object',
+                'factory() could not find object in seed',
```
mvorisek, line 145: @DarkSide666 

image

DarkSide666: ah i see. ok. if object===null, then it will throw exception below anyway.

PR #182, opened 2020-04-28
Sort numerical keys in mergeSeeds

mvorisek:


src/FactoryTrait.php
```diff
@@ -82,18 +82,33 @@ public function mergeSeeds($seed, $seed2, ...$more_seeds)
...
+                    $res[$k] = $v;
+                }
+            } elseif ($v !== null) {
+                if (!isset($res2[$k])) {
+                    $res2[$k] = $v;
+                }
             }
         }
+        ksort($res, SORT_NUMERIC);
+        foreach ($res2 as $k => $v) {
```
DarkSide666, line 107: $res = $res + $res2 ?

src/FactoryTrait.php
```diff
@@ -82,18 +82,33 @@ public function mergeSeeds($seed, $seed2, ...$more_seeds)
         }
 
         // merge seeds but prefer seed over seed2
-        foreach ($seed as $key => $value) {
-            if ($value === null && !is_numeric($key)) {
-                unset($seed[$key]);
+        // move numerical keys to the beginning and sort them
+        $res = [];
```
DarkSide666, line 86: $res = array_filter($seed, 'is_numeric', ARRAY_FILTER_USE_KEY); // with numeric keys
$seed = array_diff_key($seed, $res); // with string keys
$res2 = array_filter($seed, 'is_NOT_null, ARRAY_FILTER_USE_VALUE); // not sure howto write this properly

mvorisek: No, there is a merging logic and I think this will be also faster.

src/FactoryTrait.php
mvorisek: Why not, no collisions there :)

PR #187, opened 2020-04-30
Allow to detect/obtain breaker from hook dispatcher

### LINKED ISSUE #138: Hook break can not be detected
See https://github.com/atk4/core/blob/2d72fedda6604b0f28f9c128acc506f0a5993f23/src/HookTrait.php#L156

Currently, if breaker returns an array the break by hook breaker can not be detected.

Why not to throw an exception? If circuit breaker is desired, it should alway return an array and reference parameter like `bool &$wasInterrupted` should be added to the hook execution method. Or return an CircuitBreaker object?

mvorisek:
fix #138

### DIFF
--- a/src/HookBreaker.php
+++ b/src/HookBreaker.php
@@ -12,14 +12,22 @@ class HookBreaker extends Exception
     /**
      * @var mixed
      */
-    public $return_value;
+    protected $returnValue;
 
     /**
-     * @param mixed $return_value
+     * @param mixed $returnValue
      */
-    public function __construct($return_value)
+    public function __construct($returnValue)
     {
         parent::__construct();
-        $this->return_value = $return_value;
+        $this->returnValue = $returnValue;
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getReturnValue()
+    {
+        return $this->returnValue;
     }
 }
--- a/src/HookTrait.php
+++ b/src/HookTrait.php
@@ -163,8 +163,10 @@ trait HookTrait
      *
      * @return mixed Array of responses indexed by hook indexes or value specified to breakHook
      */
-    public function hook(string $spot, array $args = [])
+    public function hook(string $spot, array $args = [], HookBreaker &$brokenBy = null)
     {
+        $brokenBy = null;
+
         $return = [];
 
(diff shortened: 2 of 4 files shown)


src/HookBreaker.php
```diff
@@ -12,14 +12,22 @@ class HookBreaker extends Exception
     /**
      * @var mixed
      */
-    public $return_value;
+    protected $returnValue;
```
mvorisek, line 15: Full BC as `HookBreaker` was previously never returned.

src/HookTrait.php
```diff
@@ -173,24 +175,20 @@ public function hook(string $spot, array $args = [])
...
-                        $return[$index] = call_user_func_array(
-                            $data[0],
-                            array_merge(
-                                [$this],
-                                $args,
-                                $data[1]
-                            )
-                        );
+                    foreach ($_data as $index => $data) {
+                        $return[$index] = $data[0](...array_merge(
```
DarkSide666, line 179: Isn't it possible that $data[0] is defined in array format like `[$this, 'methodName']` ?

src/SessionTrait.php
```diff
@@ -69,7 +69,7 @@ public function memorize(string $key, $value)
         $this->startSession();
 
         if (is_callable($value)) {
-            $value = call_user_func($value, $key);
+            $value = $value($key);
```
DarkSide666, line 72: Isn't it possible that $value is defined in array format like `[$this, 'methodName']` ?

src/SessionTrait.php
```diff
@@ -109,7 +109,7 @@ public function recall(string $key, $default = null)
             || $_SESSION[$this->session_key][$this->name][$key] === null
         ) {
             if (is_callable($default)) {
-                $default = call_user_func($default, $key);
+                $default = $default($key);
```
DarkSide666, line 112: same here

src/HookTrait.php
mvorisek: https://3v4l.org/E5cZ5

DarkSide666: Oh perfect then!
I see my old-school php 5.x habits here :D

PR #191, opened 2020-05-07
ConfigTrait - Fix format unknown not throwing exception

abbadon1334:
If config format is unknown, the config array will be empty, without throwing an error.


DarkSide666:
Good addition, but please add one simple test case to cover this piece of code.

abbadon1334:
> Good addition, but please add one simple test case to cover this piece of code.

You are right, just add the PR and forgot.

PR #192, opened 2020-05-09
Hintable prop and method

mvorisek:
Sometimes it is needed to refer to:
- a class property - to be used a lot in data/Model
- a method name
- a method Closure - used a lot for hooks.

This PR adds a full support to it in a refactorable way. Either trait can be added to a most commonly used classes or it can be used thru static classes/methods.

Example:
```
class Cl
{
    use PropTrait;
    public $firstName;
}

$cl = new Cl(); // some object
$cl->propname()->firstName; // returns "firstName" string, but any IDE thinks
                            // "firstName" property has been used thus refactorable

// also supported for classes without PropTrait
class Address extend \stdClass {
    public $firstName;
}

$cl = new Address();
/* @var $clProp Address */
$clProp = Prop::propName($cl);
$clProp->firstName; // returns "firstName" string

// or even without an instance
/* @var $clProp */
$clProp = Prop::propName(Address::class);
$clProp->firstName; // returns "firstName" string
```

### DIFF
--- a/src/HookBreaker.php
+++ b/src/HookBreaker.php
@@ -12,14 +12,22 @@ class HookBreaker extends Exception
     /**
      * @var mixed
      */
-    public $return_value;
+    protected $returnValue;
 
     /**
-     * @param mixed $return_value
+     * @param mixed $returnValue
      */
-    public function __construct($return_value)
+    public function __construct($returnValue)
     {
         parent::__construct();
-        $this->return_value = $return_value;
+        $this->returnValue = $returnValue;
+    }
+
+    /**
+     * @return mixed
+     */
+    public function getReturnValue()
+    {
+        return $this->returnValue;
     }
 }
--- a/src/HookTrait.php
+++ b/src/HookTrait.php
@@ -163,8 +163,10 @@ trait HookTrait
      *
      * @return mixed Array of responses indexed by hook indexes or value specified to breakHook
      */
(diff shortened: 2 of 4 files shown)


mvorisek:
RTM

DarkSide666:
To me this looks like a quite big added complexity without big overall improvement.
Maybe that's because I mostly use Notepad and Sublime Text as IDEs and are used to them :)

mvorisek:
This is perfect example of feature that I consider as a MUST for larger projects, but on the other side - as also presented now - it is absolutely separateable. So based on your reactions I may without any limitation host this in a repo on my own, better?

DarkSide666:
Well, I just want to hear Romans opinion on this.

romaninsh:
So some kind of a object proxy that also makes use of reflections.. Still trying to understand what this does... @mvorisek - I'm really interested in some use-cases.

mvorisek:
@romaninsh Usecase in `$cl->propName()->firstName;` in the example

mvorisek:
@romaninsh please delete this PR from GH as there is problem with licensing

mvorisek:
@romaninsh can you delete this too?

DarkSide666:
I doubt it's possible to delete info about this PR. Code changes are no more there so I think it's fine. Right?

PR #196, opened 2020-05-14
Require closures for callbacks

mvorisek:
merge after https://github.com/atk4/data/pull/597

Callables should have context which is not the case if passes as an array. Now accept only Closure instances.

HOW TO UPGRADE:
calls with instance/array callables need to be converted to `\Closure` before adding, so calls like these:
```
$this->onHook('beforeSave', $this);
$this->onHook('beforeSave',[$this, 'beforeSave']);
```
need to be updated to:
```
$this->onHook('beforeSave', \Closure::fromCallable([$this, 'beforeSave']));
```

anonymous functions are `\Closures`, so adding them is fully supported and nothing neeeds to be changed

### DIFF
--- a/docs/dynamicmethod.rst
+++ b/docs/dynamicmethod.rst
@@ -55,7 +55,7 @@ Methods
     Tries to call dynamic method, but doesn't throw exception if it is not
     possible.
 
-.. php:method:: addMethod($name, $callable)
+.. php:method:: addMethod($name, $closure)
 
     Add new method for this object.
     See examples above.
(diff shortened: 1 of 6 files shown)


DarkSide666:
What's the point of introducing this BC breaking change?
How was `callable`s bad? Also dropped support for passing array of hook names.

mvorisek:
Yes, callables are historical concept of callbacks from PHP without context. Beside that callables are not supported for typed properties.

Standard anonymous function created like `function() {}` are `\Closures` by default, so full BC, to pass an object methods you should use `\Closure::fromCallable()` to before calling this method.

DarkSide666:
Well, probably you're right. I guess I stick to old-school style to much :)

mvorisek:
I will then fix the other main repos.

DarkSide666:
Waiting on @romaninsh . Would like to hear something from him too.

PR #197, opened 2020-05-14
mergeSeeds must reject arguments if already instanced

### LINKED ISSUE #190: mergeSeeds ignores arguments if already instanced
see https://github.com/atk4/core/blob/5fb3fbe55684f7a1c63389d021750507293668d5/src/FactoryTrait.php#L37
`$arguments` are ignored. Of couse, they can not be passed to constructor anymore as already instanced, but shouln't this result in an exception?

mvorisek:
fix #190

no BC break in most cases (no single change required for whole ui repo)

rebase/implement after https://github.com/atk4/core/pull/221

### DIFF
--- a/src/FactoryTrait.php
+++ b/src/FactoryTrait.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace atk4\core;
 
+/**
+ * @deprecated will be removed in 2021-jun
+ */
 trait FactoryTrait
 {
     /**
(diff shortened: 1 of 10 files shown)


mvorisek:
@romaninsh @DarkSide666 Please help with the tests, the `testMerge2` seems to be wrong as the seed is not an injection but array of constructor arguments. The same for `testMerge7`.

src/FactoryTrait.php
```diff
@@ -33,19 +33,36 @@ public function mergeSeeds($seed, $seed2, ...$more_seeds)
             $seed2 = $this->mergeSeeds($seed2, ...$more_seeds);
         }
 
-        if (is_object($seed)) {
+        if (is_object($seed) || is_object($seed2)) {
+            if (is_object($seed)) {
+                $passively = true; // set defaults but don't override existing properties
+            } else {
+                $passively = false;
+                [$seed, $seed2] = [$seed2, $seed]; // swap seeds
```
DarkSide666: This is clever line. I always used $tmp variable for that, but this way it's much cleaner.

DarkSide666 (commented):
Theoretically looks OK. Need to check tests and get response from @romaninsh

mvorisek:
@DarkSide666 can you fix the tests as commented above?

Once the tests are passing I would normally merge this and fix the other repos. We are on develop. We can always rollback if needed.

Did you come to a solution with @romaninsh to give me a admin merge access for these changes?

Another example is ui branch now. It needs a CS fix. As someone who normally does improve this code, I think this request is reasonable and it will help everyone.

DarkSide666:
Actually idea before was that if at least one object is passed as seed, then it should return that object and inject seed values. That's why tests fail.
```
        // object takes precedence
        $o = new SeedDITestMock();
        $this->assertSame(
            $o,
            $this->mergeSeeds([null, 'a2', null], 'b1', ['c1', null, 'c3'], $o)
        );
```
Changing this behaviour can lead to issues in other repos and projects :(

mvorisek:
Yes and no, no because the arguments were previously ignored if object was passed.

> Actually idea before was that if at least one object is passed as seed, then it should return that object and inject seed values. That's why tests fail.

> inject seed values

but there are only arguments instead of key-value pairs to inject :) - previously the seed was accumulated but never set to a seed object once instanced.

mvorisek:
@DarkSide666 can you fix/remove the failing tests so we can merge and see/fix the impact on other repos?

romaninsh (approved):
that would expose some previously ignored hidden bugs.

PR #198, opened 2020-05-14
Unify PHP config - always require return

mvorisek:
Safe, if there is no return, config is not an array, thus an exception is thrown.


mvorisek:
@abbadon1334 ok?

docs/config.rst
```diff
@@ -35,18 +35,6 @@ Supported Formats
 
 php
 ---
-Configuration is declared in a variable $config
-
-    $config=[
```
DarkSide666, line 40: @romaninsh This was quite common syntax of config.php file in our old projects... but I guess it's better to return values and we can assign them to variable of out choice.

PR #199, opened 2020-05-15
Use MB functions

### LINKED ISSUE #27: Use mb_ functions
Check if we have any of these methods and if so,then use their mb_ equivalent.
http://php.net/manual/en/ref.mbstring.php

Particularly test:
- [ ] passing utf8 values as parameters paramns from exceptions
- ..

DarkSide666:
fix #27

### DIFF
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -51,7 +51,6 @@ return PhpCsFixer\Config::create()
         'nullable_type_declaration_for_default_null_value' => [
             'use_nullable_type_declaration' => false,
         ],
-        'mb_str_functions' => true,
     ])
     ->setFinder($finder)
     ->setCacheFile(__DIR__ . '/.php_cs.cache');
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -51,6 +51,7 @@ return PhpCsFixer\Config::create()
         'nullable_type_declaration_for_default_null_value' => [
             'use_nullable_type_declaration' => false,
         ],
+        'mb_str_functions' => true,
     ])
     ->setFinder($finder)
     ->setCacheFile(__DIR__ . '/.php_cs.cache');
--- a/composer.json
+++ b/composer.json
@@ -37,6 +37,7 @@
   "require": {
     "php": ">=7.2.0",
     "ext-json": "*",
+    "ext-mbstring": "*",
     "psr/log": "~1.0",
     "symfony/polyfill-php73": "^1.15",
     "symfony/yaml": ">=3.4"
--- a/composer.json
+++ b/composer.json
@@ -37,7 +37,6 @@
   "require": {
     "php": ">=7.2.0",
     "ext-json": "*",
-    "ext-mbstring": "*",
     "psr/log": "~1.0",
     "symfony/polyfill-php73": "^1.15",
     "symfony/yaml": ">=3.4"
--- a/src/CollectionTrait.php
+++ b/src/CollectionTrait.php
@@ -164,7 +164,7 @@ trait CollectionTrait
         if (
             isset($this->_appScopeTrait) &&
             isset($this->app->max_name_length) &&
-            strlen($desired) > $this->app->max_name_length
+            mb_strlen($desired) > $this->app->max_name_length
         ) {
             /*
              * Basic rules: hash is 10 character long (8+2 for separator)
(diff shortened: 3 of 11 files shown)


.php_cs.dist
```diff
@@ -51,6 +51,7 @@ return PhpCsFixer\Config::create()
         'nullable_type_declaration_for_default_null_value' => [
             'use_nullable_type_declaration' => false,
         ],
+        'mb_str_functions' => true,
```
mvorisek: Not all string functions should be migrated, make sure this is removed before merge!

composer.json
```diff
@@ -37,6 +37,7 @@
   "require": {
     "php": ">=7.2.0",
     "ext-json": "*",
+    "ext-mbstring": "*",
```
mvorisek: As far as I know always presented in PHP, at least in the newer versions (and also a discussion in PHP group to make json presented as well in PHP 8.0)

src/CollectionTrait.php
```diff
@@ -164,19 +164,19 @@ protected function _shorten_ml(string $desired): string
         if (
             isset($this->_appScopeTrait) &&
             isset($this->app->max_name_length) &&
-            strlen($desired) > $this->app->max_name_length
+            mb_strlen($desired) > $this->app->max_name_length
```
mvorisek, line 167: is `max_name_length` for UTF-8 size? (I would say not for the indended behaviour - like shortening keys for browser, but I think it is ok)

Also note, that mb_xxx functions are about 10x slower.

src/ConfigTrait.php
```diff
@@ -59,7 +59,7 @@ public function readConfig($files = ['config.php'], string $format = 'php')
 
             $tempConfig = [];
 
-            switch (strtolower($format)) {
+            switch (mb_strtolower($format)) {
```
mvorisek: I would not write this normally using mb_xxx. But ok to keep...

src/FactoryTrait.php
```diff
@@ -226,14 +226,14 @@ public function normalizeClassName(string $name, string $prefix = null): string
...
-            $name = $prefix . '\\' . substr($name, 1);
+            $name = $prefix . '\\' . mb_substr($name, 1);
             $name = str_replace('/', '\\', $name);
 
             return $name;
         }
 
         // Rule 2: if "\" is present, don't prefix
-        if (strpos($name, '\\') !== false) {
+        if (mb_strpos($name, '\\') !== false) {
```
mvorisek: is is totally not needed, only slower...

mvorisek:
@DarkSide666 I reverted some changes that does not make sense to me.

Do you have an idea how to deduplicate the code in `ContainerTrait::_shorten` and `CollectionTrait::_shorten_ml`? It is copy/paste... Hate it!

.php_cs.dist
DarkSide666: Why not all?

composer.json
DarkSide666: ok this can be removed if you say so

src/FactoryTrait.php
DarkSide666: oh yes you're right

src/ConfigTrait.php
DarkSide666: format will never be multibyte so probably this also needs to be removed.

DarkSide666:
> Do you have an idea how to deduplicate the code in `ContainerTrait::_shorten` and `CollectionTrait::_shorten_ml`? It is copy/paste... Hate it!

Only option I see is to move this method to new "parent" Trait, but traits can't be extended so we will always need to add this other trait in classes. That will not be nice solution.

.php_cs.dist
mvorisek: Why all?

We can migrate all text functions as long as:
- it makes sense (like `strpos('ascii char')`, `strtoupper($http_method)`, strtoupper('hex string') are great examples where is does not)
- if the performance is not critical (like for magic name getters)
...

When never:
- if the string is binary - like checking lenght of binary file (ex. png image)
- when the lenght is checked for some storage limitation - a lot of applications limit the bytes count instead of Unicode symbols/chars.

mvorisek:
They can be extended - see https://stackoverflow.com/questions/40299080/is-there-a-way-to-extend-a-trait-in-php

.php_cs.dist
DarkSide666: OK I rolled back part of changes in other repos too. Please review them when you have time.

src/CollectionTrait.php
DarkSide666: well... names can theoretically be anything so can contain utf8 chars.

DarkSide666:
> They can be extended - see https://stackoverflow.com/questions/40299080/is-there-a-way-to-extend-a-trait-in-php

That's great! Then we maybe can create `Helper` trait which would contain method which can be used across multiple traits and implement it like https://stackoverflow.com/a/40299247/1466341

georgehristov:
Just a note: Trait in Trait duplication works only in PHP >=7.3

src/CollectionTrait.php
mvorisek: I would bet the hard limitations like in suhosin (I do not use it) are for bytes, but I support modern tech as much as possible, to I am perfectly fine with this

DarkSide666:
Then it will be one more thing in our to do list when we drop php 7.2

mvorisek:
As far as I know they can be aliased so conflicts can be resolved for any PHP 7.

PR #202, opened 2020-05-16
WIP: Do not pass this by default

### LINKED ISSUE #201: It should be possible to add hooks that do not require the source object as 1st param
Example of bad but typical usage:
image

Example of better but still not good usage:
image

Example of good usage:
image

mvorisek:
Fix #201

Once #196 is merged the arguments from `onHook` should be not longer needed at all.

### DIFF
--- a/src/HookArgThis.php
+++ b/src/HookArgThis.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace atk4\core;
+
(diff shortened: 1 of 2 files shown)


mvorisek:
Adding "in stand by" by myself, probably too soon

mvorisek:
Maybe later or I need to come up with a solution that supports cloning/rebounding...

PR #203, opened 2020-05-16
Delete HTMLText exception renderer class

mvorisek:
merge after: https://github.com/atk4/ui/pull/1246

This PR removes `Exception::getHTMLText()` method and `ExceptionRenderer/HTMLText` renderer. Native HTML renderer should be used instead.

The only use is here: https://ui.agiletoolkit.org/demos/interactive/console.php


abbadon1334:
> What is the usecase of this?
> 
> Can we use some industry standard convertor to convert console/colored str to HTML or HTML to text?

I don't remember the use case of "simplified html/text", but i think is for integration in others libraries.

In this specific case, when Exception is throw we don't need performance, is the end of a bad request.

i don't know if is correct to added some require to composer only for marginal functionalities, like : https://github.com/mtibben/html2text

every Renderer, process and return the processed exception without echo something, it doesn't break output.

how to know if someone use it and we can remove without break something somewhere?

My idea was working in direction of making atk4 usable with PSR7-PSR15 to be integrated more easy everywhere.

mvorisek:
Closing, the whole renderer concept should be reworked to address more than this issue.

mvorisek:
It seems it is used for only in `\atk4\ui\Console` - we should slighly the modify the HTML renderer instead and delete this. Hard to maintain n versions, really bad design.

romaninsh:
@mvorisek so what did you decide to do at the end? Could you update PR description / title accordingly? 

I think we should keep HTML rendering for output in consoles, but code can relocate as necessary. I don't think it needs to be changed much though, so not hard to support.

mvorisek:
> so what did you decide to do at the end? Could you update PR description / title accordingly?

@romaninsh to delete :), title ok

pure console renderer for CLI remains

> I think we should keep HTML rendering for output in consoles, but code can relocate as necessary. I don't think it needs to be changed much though, so not hard to support.

it is, currently the renderers are not very well written as they can not be extended further, are for single use only etc.

now the exceptions in console are rendered like: https://user-images.githubusercontent.com/2228672/83938850-59d30e80-a7d8-11ea-85f3-e196d09f0cb5.png much, much better UX

DarkSide666 (approved):
LGTM

PR #205, opened 2020-05-16
Validate hook closure signatures

mvorisek:
Finish after #196

~related with #202~

should be fully BC


mvorisek:
maybe in the future when we will define stub for spot instead in the calling side, the idea is probably too strict as it violates Liskov principe

PR #211, opened 2020-06-01
Do not normalize "/" in class names

mvorisek:
Proper usage should be fully BC. Usage of `/` as NS separator will throw an error.

Normalizing "/" to "\\" in class names does not add any advantage.

The current implementation breaks anonymous class names.

Example:
```
$seed = [get_class(new class () extends \atk4\ui\FormField\Upload {
    ...
})]
```
is valid seed and this PR fixes it.

### DIFF
--- a/src/FactoryTrait.php
+++ b/src/FactoryTrait.php
@@ -217,7 +217,6 @@ trait FactoryTrait
             && method_exists($this->app, 'normalizeClassNameApp')
         ) {
             $result = $this->app->normalizeClassNameApp($name, $prefix);
-
             if ($result !== null) {
                 return $result;
             }
--- a/src/FactoryTrait.php
+++ b/src/FactoryTrait.php
@@ -191,14 +191,14 @@ trait FactoryTrait
      * add prefix.
      *
      * Rules observed, in order:
-     *  - If class starts with "./" then prefixing is always done.
+     *  - If class starts with ".\" then prefixing is always done.
      *  - If class contains "\" or it is an anonymous class prefixing is never done.
      *  - If class (with prefix) exists, do prefix.
      *  - don't prefix otherwise.
      *
      * Example: normalizeClassName('User', 'Model') == 'Model\User';
      * Example: normalizeClassName(Test\User::class, 'Model') == 'Test\User'; # or as per "use"
-     * Example: normalizeClassName('./User', 'Model') == 'Model\User';
+     * Example: normalizeClassName('.\User', 'Model') == 'Model\User';
      * Example: normalizeClassName('User', 'Model') == 'Model\User'; # if exists, 'User' otherwise
      *
      * # If used without namespace:
@@ -222,12 +222,12 @@ trait FactoryTrait
             }
         }
 
-        // Rule 1: if starts with "./" always prefix
-        if (strpos($name, './') === 0 && $prefix) {
-            return $prefix . '\\' . substr($name, 2);
+        // Rule 1: if starts with ".\" always prefix
+        if (strpos($name, '.\\') === 0 && $prefix) {
+            return $prefix . substr($name, 1);
         }
 
-        // Rule 2: if "\" is present, don't prefix
+        // Rule 2: if contains "\" never prefix
         if (strpos($name, '\\') !== false || strpos($name, 'class@anonymous') === 0) {
             return $name;
         }
--- a/src/FactoryTrait.php
+++ b/src/FactoryTrait.php
@@ -191,14 +191,13 @@ trait FactoryTrait
      * add prefix.
      *
      * Rules observed, in order:
-     *  - If class starts with "." then prefixing is always done.
-     *  - If class contains "\" prefixing is never done.
+     *  - If class starts with "./" then prefixing is always done.
+     *  - If class contains "\" or it is an anonymous class prefixing is never done.
      *  - If class (with prefix) exists, do prefix.
      *  - don't prefix otherwise.
      *
      * Example: normalizeClassName('User', 'Model') == 'Model\User';
      * Example: normalizeClassName(Test\User::class, 'Model') == 'Test\User'; # or as per "use"
-     * Example: normalizeClassName('Test/User', 'Model') == 'Model\Test\User';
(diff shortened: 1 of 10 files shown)


src/FactoryTrait.php
```diff
@@ -191,14 +191,13 @@ public function factory($seed, $defaults = [], string $prefix = null): object
...
+     *  - If class starts with "./" then prefixing is always done.
+     *  - If class contains "\" or it is an anonymous class prefixing is never done.
      *  - If class (with prefix) exists, do prefix.
      *  - don't prefix otherwise.
      *
      * Example: normalizeClassName('User', 'Model') == 'Model\User';
      * Example: normalizeClassName(Test\User::class, 'Model') == 'Test\User'; # or as per "use"
-     * Example: normalizeClassName('Test/User', 'Model') == 'Model\Test\User';
      * Example: normalizeClassName('./User', 'Model') == 'Model\User';
      * Example: normalizeClassName('User', 'Model') == 'Model\User'; # if exists, 'User' otherwise
```
DarkSide666, line 202: Comment # if exists, User otherwise is wrong here. I do not see any class_exists() method call here.

src/FactoryTrait.php
```diff
@@ -191,14 +191,13 @@ public function factory($seed, $defaults = [], string $prefix = null): object
...
-     *  - If class starts with "." then prefixing is always done.
-     *  - If class contains "\" prefixing is never done.
+     *  - If class starts with "./" then prefixing is always done.
+     *  - If class contains "\" or it is an anonymous class prefixing is never done.
      *  - If class (with prefix) exists, do prefix.
      *  - don't prefix otherwise.
      *
      * Example: normalizeClassName('User', 'Model') == 'Model\User';
      * Example: normalizeClassName(Test\User::class, 'Model') == 'Test\User'; # or as per "use"
-     * Example: normalizeClassName('Test/User', 'Model') == 'Model\Test\User';
```
DarkSide666, line 201: If this syntax with ```/``` is no more supported, then this PR is BC.

src/FactoryTrait.php
```diff
@@ -191,14 +191,13 @@ public function factory($seed, $defaults = [], string $prefix = null): object
...
      * Example: normalizeClassName('./User', 'Model') == 'Model\User';
```
DarkSide666, line 201: What if we have `.\User` here?

src/FactoryTrait.php
mvorisek: ?

src/FactoryTrait.php
mvorisek: Yes, if anyone uses wrong names...

src/FactoryTrait.php
mvorisek: I do not like two variants or relative paths at all. But as long as relative paths are allowed, probably both variants can be allowed.

As the NS separator is `\\`, allow only `.\\` instead of `./`?

mvorisek: fixed, good?

src/FactoryTrait.php
mvorisek: fixed in desc, added label

mvorisek: is it ok to add `bug` label on PRs which fixes some bug?

src/FactoryTrait.php
DarkSide666: That comment at the end of this line. We do not check for class existence in this method, just normalize class name.

mvorisek: You mean `or as per "use"`? I left this untouched, I basically removed `\` support only.

src/FactoryTrait.php
```diff
@@ -196,13 +196,11 @@ public function factory($seed, $defaults = [], string $prefix = null): object
...
      *  - don't prefix otherwise.
      *
-     * Example: normalizeClassName('User', 'Model') == 'Model\User';
-     * Example: normalizeClassName(Test\User::class, 'Model') == 'Test\User'; # or as per "use"
-     * Example: normalizeClassName('.\User', 'Model') == 'Model\User';
-     * Example: normalizeClassName('User', 'Model') == 'Model\User'; # if exists, 'User' otherwise
-     *
-     * # If used without namespace:
-     * Example: normalizeClassName(User::class, 'Model') == 'Model\User'; # if exists, 'User' otherwise
+     * Example: normalizeClassName('User') == 'User'; # default -> no prefix
```
mvorisek, line 199: `normalizeClassName('User') -> 'User'` seems better to me, rules are above, more comments, more distractions...

src/FactoryTrait.php
DarkSide666: I know a project which contains hundreds of classnames like `namespace/modelName` :)
But you're right, it's wrong slash there. I was just wondering maybe we still could have `str_replace('/', '\\', $name)` there at the beginning and then it would be fully compatible.
In this particular case I do not see big advantage of being so strict.

mvorisek: I think we should enforce the proper way. The replace was breaking at least anonymous classes, later, this prefixing should be completely removed, thus we should enforce the right usage now.

DarkSide666: well, ok :)

src/FactoryTrait.php
```diff
@@ -220,17 +222,16 @@ public function normalizeClassName(string $name, string $prefix = null): string
...
-        // Rule 1: if starts with ".\" then always prefix
-        if (strpos($name, '.\\') === 0 && $prefix) {
-            return $prefix . substr($name, 1);
-        }
-
-        // Rule 2: if contains "\" or is anonymous class then never prefix
-        if (strpos($name, '\\') !== false || strpos($name, 'class@anonymous') === 0) {
-            return $name;
+        if ($prefix !== null && (substr($prefix, 0, 1) === '\\' || substr($prefix, -1) === '\\')) {
+            throw new Exception('Class prefix must not start/end with backslash');
```
DarkSide666, line 226: Why not just trim slashes instead?

mvorisek: To enforce meaningfull/strict input.

PR #217, opened 2020-06-03
Add support of runtime type checking and type checked factory

mvorisek:
Indended to replace a lot of subtype check in data/ui, where is it code in many ways and not even checked in a lot of placed

then merge immediatelly: https://github.com/atk4/ui/pull/1237


romaninsh:
@mvorisek could you please expand on the description (title and body)?

mvorisek:
> could you please expand on the description (title and body)?

done, RTM

DarkSide666 (approved):
ok

PR #219, opened 2020-06-04
Better factory prefix detection/throw

mvorisek:
zero change except throwing `\Error` which should be fully caught/supported by ui exception handler/renderer


mvorisek:
will be merged in https://github.com/atk4/core/pull/221

PR #220, opened 2020-06-04
Always build html safe name

mvorisek:
Needed at least for anonymous classes at their names contain unsafe characters (`\0` etc.)

Related with (now reverted) PR: https://github.com/atk4/ui/pull/1230

Marking as BC breaking as the generated named might be used by someone in static CSS rules

### DIFF
--- a/src/TrackableTrait.php
+++ b/src/TrackableTrait.php
@@ -52,7 +52,7 @@ trait TrackableTrait
             $name .= '@anonymous';
         }
 
-        return trim(preg_replace('~^atk4\\\\[^\\\\]+\\\\|[^0-9a-z_]+~is', '_', mb_strtolower($name)), '_');
+        return trim(preg_replace('~^atk4\\\\[^\\\\]+\\\\|[^0-9a-z\x7f-\xfe]+~s', '_', mb_strtolower($name)), '_');
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/TrackableTrait.php
```diff
@@ -40,7 +40,19 @@ trait TrackableTrait
      */
     public function getDesiredName(): string
     {
-        return preg_replace('/.*\\\\/', '', mb_strtolower(static::class));
+        // can be anything, but better to build meaningful name
+        $name = static::class;
+        if (strpos($name, 'class@anonymous') === 0) {
+            $name = '';
+            foreach (class_parents(static::class) as $name) {
```
PhilippGrashoff: I do not exactly figure this foreach loop. To my understanding, `$name` defined in `as $name` should be only available inside the foreach loop and have nothing to do with the outer `$name` var. Or am I wrong here and the outer `$name` is changed within this foreach? Because If not I dont see what the foreach is doing

mvorisek: I improved it now, it worked but was not that clean. (and it was actually wrong for nested anonymous classes, but they are not supported by PHP as they can not be created by the current PHP grammar)

ibelar (approved):
LGTM

PR #221, opened 2020-06-04
Add central/singleton Factory and deprecate FactoryTrait trait

mvorisek:
no BC break

merge after: https://github.com/atk4/ui/pull/1251 (one line fix needed for ui/Template because of php7.2 limitation of traits)

By using single point factory we can offer extreme magic - user can inject own Factory and he can return any implementation he wants for any object (based on stacktrace even different) thru the whole project.

For these reasons you should never use direct instantiation:
```
new Class($arg1, $args2, ...)
```
in Atk code but always create new instance thru `\atk4\core\Factory` like:
```
Factory::factory([Class::class, $arg1, $rg2, ...])
```

FactoryTrait was quite a bad design pattern - there is currently no need to remove it, as it can be deprecated only for now without any negative side effect.

### DIFF
--- a/src/DIContainerTrait.php
+++ b/src/DIContainerTrait.php
@@ -100,7 +100,8 @@ trait DIContainerTrait
     public static function assertInstanceOf(object $object)// :static supported by PHP8+
     {
         if (!($object instanceof static)) {
-            throw (new Exception('Object is not an instance of ' . static::class))
+            throw (new Exception('Object is not an instance of static class'))
+                ->addMoreInfo('static_class', static::class)
                 ->addMoreInfo('object_class', get_class($object));
         }
 
@@ -125,7 +126,8 @@ trait DIContainerTrait
 
             $cl = $seed[0];
             if (!$unsafe && !is_a($cl, static::class, true)) {
-                throw (new Exception('Seed class is not a subtype of ' . static::class))
+                throw (new Exception('Seed class is not a subtype of static class'))
+                    ->addMoreInfo('static_class', static::class)
                     ->addMoreInfo('seed_class', $cl);
             }
         }
--- a/src/DIContainerTrait.php
+++ b/src/DIContainerTrait.php
@@ -121,7 +121,8 @@ trait DIContainerTrait
             }
 
             if (!isset($seed[0])) {
-                throw (new Exception('Class name is not defined in seed'));
+                throw (new Exception('Class name is not specified by the seed'))
+                    ->addMoreInfo('seed', $seed);
             }
 
             $cl = $seed[0];
--- a/src/DIContainerTrait.php
+++ b/src/DIContainerTrait.php
@@ -88,4 +88,91 @@ trait DIContainerTrait
             ->addMoreInfo('property', $key)
             ->addMoreInfo('value', $value);
     }
+
+    /**
+     * Return the argument and assert it is instance of current class.
+     *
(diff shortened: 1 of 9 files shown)


src/Core.php
```diff
@@ -0,0 +1,41 @@
...
+     */
+    public static function mergeSeeds(...$seeds)
+    {
+        return self::getFactory()->mergeSeeds(...$seeds);
+    }
+
+    /**
+     * See \atk4\core\Factory::factory().
+     */
+    public static function factory($seed, $defaults = [], /* remove in 2021-jun, catch passed but no longer supported prefix */...$extraArgs): object
```
georgehristov: How about renaming this `create` for semantic purposes?

src/Core.php
```diff
@@ -0,0 +1,41 @@
...
+            self::$_factory = new Factory();
+        }
+
+        return self::$_factory;
+    }
+
+    /**
+     * See \atk4\core\Factory::mergeSeeds().
+     */
+    public static function mergeSeeds(...$seeds)
```
georgehristov: How about using only `merge`?

DarkSide666:
This one I definitely would like @romaninsh to review too.

romaninsh (approved):
Agree that this is not a BC, however updating other core classes to use new Factory trait would cause BC..

It's better if we plan it and suggest users how to update their code.

src/DIContainerTrait.php
```diff
@@ -88,4 +88,94 @@ protected function setMissingProperty($key, $value)
...
+     * The best, typehinting-friendly, way to create an object if it should not be
+     * immediately added to a parent (otherwise use addTo() method).
+     *
+     * @param array|string|object $seed the first element specifies a class name, other elements are seed
+     *
+     * @return static
+     */
+    public static function fromSeed($seed = [], $defaults = [])// :static supported by PHP8+
+    {
+        if (func_num_args() > 2) { // prevent bad usage
```
georgehristov, line 151: You can reduce code here by simpy calling
```
$object = static::fromSeedUnsafe($seed, $defaults);

static::assertInstanceOf($object);

return $object;
```

mvorisek: ⚠️ Can, but do not want - this way, `assertInstanceOf` is computed even before the object is even created, which implies stronger security! (as seed can be an user input)

src/Core.php
mvorisek: I am not against, but I would like to see analysis from top 10 frameworks to decise if it is worth or not.

src/Core.php
mvorisek: Current name is much more self explaning, I prefer to keep it.

src/DIContainerTrait.php
georgehristov: What I see the `assertInstanceOf` is called before the `$object` is returned only?
The rest of the code is the same.

mvorisek: ⚠️ No! `_fromSeedPrecheck` is called BEFORE object is constructed. Be very carefull around this and test for this behaviour will be highly welcomed.

src/Core.php
mvorisek: You should always use `ClWithDi::fromSeed`, thus renaming this method does not help with daily usage...

src/DIContainerTrait.php
mvorisek: PR for test: https://github.com/atk4/core/pull/311

PR #222, opened 2020-06-04
Display exception location as another stack trace call

### LINKED ISSUE #213: Exception location is not dumped
image

mvorisek:
fixes #213

Example of this PR in action:
```
try {
  (function() {
        throw new \atk4\ui\Exception('No database connection');
  })();
} catch (\Exception $e) {
    throw new \Exception('x', 0, $e);
}
```

image


DarkSide666 (approved):
Nice

PR #223, opened 2020-06-05
Exception constructor no longer accepts array of solutions/params

mvorisek:
merge after: https://github.com/atk4/ui/pull/1242

Remove bad design of this core component. Wrong usage will throw now `\TypeError`.

Previously there were also a lot of non-sense usages like:
```
throw new Exception(['Model is not associated with any database']);
```
which only underlays how much confusion the original design was introducing.

Regex for potential places to update: `new [^\n]*?exception\(\[`

To update your code refactor all usages of Exception constructor usages with array like:
```
throw new Exception([
    'Object with requested name already exist in collection',
    'name' => $name,
    'collection' => $collection,
]);
```
to:
```
throw (new Exception('Object with requested name already exist in collection'))
    ->addMoreInfo('name', $name)
    ->addMoreInfo('collection', $collection);
```

There was also a special `solutions` keys, see PR source, used nowhere except one test in core repo - regex `new [^\n]*?exception\(\[(?:.|\n)+?'solutions'`.

you can also fix your code by script: https://github.com/mvorisek/atk4_ui_add_migrate/tree/fix_array_exception_constructor (but run CS fixer then and check manually)

### DIFF
--- a/docs/exception.rst
+++ b/docs/exception.rst
@@ -21,7 +21,7 @@ the constructor or using addMoreInfo()
     This uses same format as a regular PHP exception, but error parameter will
     now support array::
 
-        throw new Exception(['Value is too big', 'max'=>$max]);
+        throw (new Exception('Value is too big'))->addMoreInfo('max', $max);
 
 The other option is to supply error is:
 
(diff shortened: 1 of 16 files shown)


DarkSide666 (approved):
LGTM

PR #224, opened 2020-06-05
Temporary fix for ui tests - try to produce the same component names as before #220

mvorisek:
revert after: https://github.com/atk4/ui/pull/1243


ibelar (approved):
LGTM

PR #228, opened 2020-06-07
Fix too big segment div in html exception param render

mvorisek:
fixed style seems bad to me, should be probably solved like the table below with `very compact` style and both table headers should be reduced

important for huge nested exceptions

Example of current renderer:
image

all header/row heights and indentations from left/right should be unified.

help wanted


mvorisek:
@abbadon1334 can you look at this one?

mvorisek:
replaced/fixed by https://github.com/atk4/core/pull/261

PR #229, opened 2020-06-07
Fix exception renderer makeRelativePath() for empty path

mvorisek:


romaninsh (approved):
bit cryptic but not harmful.

PR #231, opened 2020-06-07
Fix composer develop install issues

### LINKED ISSUE #193: Every dev branch of atk4 repos should replace the 2.x version
See https://github.com/atk4/core/blob/313e372f30eb249ad1375f13c84acef232340a98/composer.json#L45

Important to fix all cases as mentioned by @DarkSide666 to make develop versions easily installable from custom projects with intermediate libraries, which requires otherwise stable version. Currently it is not possible to install them.

Symfony seems to be using `branch-alias` composer feature: https://github.com/twigphp/Twig/blob/3.x/composer.json#L46

mvorisek:
fixes https://github.com/atk4/core/issues/193

these two lines should stay on release (they are for root composer project only, they may still be required for dev deps):
```
  "minimum-stability": "dev",
  "prefer-stable": true,
```

but this one:
```
  "version": "dev-develop",
```
should be remove completely on release.

For all repos.

### DIFF
--- a/.github/workflows/bundler.yml
+++ b/.github/workflows/bundler.yml
@@ -16,6 +16,7 @@ jobs:
         run: |
           # replaces X keys with X-release keys
           jq '. as $in | reduce (keys_unsorted[] | select(endswith("-release")|not)) as $k ({}; . + {($k) : (($k + "-release") as $kr | $in | if has($kr) then .[$kr] else .[$k] end) } )' < composer.json > tmp && mv tmp composer.json
+          composer config version --unset
           v=$(echo ${{ github.ref }} | cut -d / -f 4)
           echo "::set-env name=version::$v"
 
--- a/composer.json
+++ b/composer.json
@@ -50,7 +50,7 @@
     "symfony/contracts": ">=1.1"
   },
   "require-dev-release": {
-    "atk4/data": "^2.0",
+    "atk4/data": "^2.2.0",
     "friendsofphp/php-cs-fixer": "^2.16",
     "phpunit/phpunit": ">=9.0",
     "symfony/contracts": ">=1.1"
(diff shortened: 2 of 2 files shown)


DarkSide666 (approved):
- in current develop branch composer updates fine
- in develop branch with version:dev-develop composer updates fine
- in another branch without version:dev-develop composer DO NOT UPDATE - atk4/data dev-develop requires atk4/core dev-develop
- in another branch with version:dev-develop composer updates fine

So it looks that adding version:dev-develop solves this issue.

DarkSide666:
If we can replace, then something like this should work already:
```
// these will be used in develop
"version": "dev-develop",
"minimum-stability": "dev",
"prefer-stable": true,

// these will be used when releasing
"version-release": "something",
"minimum-stability-release": "something",
"prefer-stable-release": true,
```

DarkSide666:
TODO - need to add another `jq` line in `core/.github/workflows/bundler.yml` to completely remove `version` key from json while releasing.

mvorisek:
`composer config version --unset` seems to be much cleaner approach

I think we can also remove `require(-dev)?-release` completely and in release process replace all `(?<=atk4/xxx: )dev-*` requires with `^version`

here is how to get a version: https://github.com/atk4/ui/blob/09ae0f67deda221ba2f48d6d3beb00be01e7a010/.github/workflows/bundler.yml#L41

mvorisek:
about deps:

we should drop `require-release` concept in composer.json and instead replace `dev-develop` with `^version`

mvorisek:
@romaninsh @DarkSide666 updated post about deps, wdyt? Very robust if we can let the release script do the work...

DarkSide666:
having `...-release` is still ok for now, so let's not complicate this at this point.

PR #232, opened 2020-06-07
Update composer.json to avoid version clashes

romaninsh:
Merge after atk4/data 2.1 is released.

## TODO

 - [ ] create identical PR for develop branch (can't re-use)

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,7 +42,7 @@
     "symfony/yaml": ">=3.4"
   },
   "require-dev": {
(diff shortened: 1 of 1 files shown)


mvorisek:
@romaninsh I think BC branches does not solve anything related with FC - we should release like every two week when major/breaking changes were done and allow microstepping instead.

With this approach, we can guaranteee full stable functionality.

For these reasons, I am closing this.

In releases, we should require exact (same) version in all major repos (released at once).

PR #233, opened 2020-06-08
Refactor hasXXX() to tryGetXXX() and change return from false to null

mvorisek:
BC maintaned, but hasXXX compatibility marked as deprecated and it should emit warning after a few releases

check related PR https://github.com/atk4/core/pull/240

### DIFF
--- a/docs/container.rst
+++ b/docs/container.rst
@@ -85,7 +85,7 @@ Example::
 
         public function tryGetField($name)
         {
-            return $this->_TryGetFromCollection($name, 'fields');
+            return $this->_tryGetFromCollection($name, 'fields');
         }
 
         public function getField($name)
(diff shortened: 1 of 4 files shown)


docs/container.rst
```diff
@@ -83,9 +83,9 @@ Example::
             return $this->_addIntoCollection($name, $field, 'fields');
         }
 
-        public function hasField($name)
+        public function tryGetField($name)
         {
-            return $this->_hasInCollection($name, 'fields');
+            return $this->_TryGetFromCollection($name, 'fields');
```
DarkSide666: First letter is uppercase?

DarkSide666 (approved):
ok, fine if we keep BC for reasonable time.

PR #235, opened 2020-06-08
Better fix for vendor path discovery

mvorisek:
fix for usecases:
- `\atk4\core` is main composer project
- `\atk4\core` is dependency
- `\Composer\Autoload\ClassLoader` is loaded from other composer (special case but supported by Composer)


romaninsh (approved):
LGTM

DarkSide666 (approved):
LGTM

PR #239, opened 2020-06-10
Remove totally unmaintained @throws phpdocs

mvorisek:
In top notch code they may help, but in Atk there are defined very randomly, better to remove all @throws phpdocs except some very specific/meaningful.

Example before this commit - what is the difference? Confusion only.
image


DarkSide666 (approved):
I agree. We should always think that basically everything can throw any exception.

PR #240, opened 2020-06-11
HasXxx() methods now return bool only

mvorisek:
Cleaner approach to https://github.com/atk4/core/pull/233

I think it is much better fix to hasXxx methods - originally I was targetting to solve to not return `object|false` type, but the whole concept was not that good - I understand, that tryGetXxx might be handy in a few cases, but the argument can be stored in var, checked for existence, and the used with getXxx. In other (most) cases, this was used to check solely for existance which we continue to support without change.

### DIFF
--- a/docs/container.rst
+++ b/docs/container.rst
@@ -245,8 +245,8 @@ Methods
 
 .. php:method:: hasElement($short_name)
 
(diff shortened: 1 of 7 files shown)


DarkSide666 (approved):
ok

PR #241, opened 2020-06-11
Remove @throws by CS fixer

mvorisek:
related with https://github.com/atk4/core/pull/239


mvorisek:
some @throws may be ok - closing

DarkSide666 (approved):

PR #243, opened 2020-06-12
Delete QuickExceptionTrait.php and PsyshE.php

mvorisek:


DarkSide666 (approved):
Removing QuickException looks fine, but not so sure about removing PsyshE.
Anyway - approved.

PR #244, opened 2020-06-12
Do not silence missing numeric property in setDefaults()

mvorisek:
ui tests passing

must be merged at once with https://github.com/atk4/ui/pull/1274

`!is_numeric($key)` check is not needed as `property_exists($this, $key)` will fail due strict types (and `$key` will be int because numeric array keys are always `int`)

### DIFF
--- a/docs/di.rst
+++ b/docs/di.rst
@@ -63,7 +63,7 @@ How to use DIContainerTrait
 
 .. php:trait: DIContainerTrait
 
-.. php:method: setDefaults($array)
+.. php:method: setDefaults($properties, $passively = false)
 
 .. php:method: setMissingProperty($propertyName, $value)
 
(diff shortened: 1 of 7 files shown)


docs/di.rst
```diff
@@ -63,9 +63,9 @@ How to use DIContainerTrait
 
 .. php:trait: DIContainerTrait
 
-.. php:method: setDefaults($array, $strict = false)
+.. php:method: setDefaults($array)
```
DarkSide666: We have `    public function setDefaults(array $properties, bool $passively = false)`

DarkSide666 (approved):

docs/di.rst
mvorisek: yeah, good one, docs should be generated one day...

PR #227, opened 2020-06-06
Remove trait identification property

mvorisek:
finish after: https://github.com/atk4/core/pull/221

then remove dummy properties like:
image

and refactor in other repos

Trait identification properties declared in `core`:
```
_appScopeTrait|_configTrait|_containerTrait|_debugTrait|_DIContainerTrait|_dynamicMethodTrait|_factoryTrait|_hookTrait|_initializerTrait|_nameTrait|_quickExceptionTrait|_sessionTrait|_staticAddToTrait|_trackableTrait|_translatableTrait
```
-> they are used solely in `core` repo except 3 usages in `ui` repo

PR #245, opened 2020-06-12
Add get/set support for DIContainerTrait::setDefaults()

### LINKED ISSUE #206: Allow/prioritize setter methods in DIContainerTrait::setDefaults()
https://github.com/atk4/core/blob/503ed2ff35865ca8b5499ecb61c95a6257b95621/src/DIContainerTrait.php#L50

proposal to call setXX() if exists instead of setting the XX property directly

mvorisek:
fixes https://github.com/atk4/core/issues/206 , needed for https://github.com/atk4/core/pull/227

done, but need tests

maybe even add/remove with syntax like `class.adds = ['x', 'y'],` ...

### DIFF
--- a/src/DiContainerTrait.php
+++ b/src/DiContainerTrait.php
@@ -66,7 +66,7 @@ trait DiContainerTrait
                     if ($origValue !== null) {
                         $isMissing = false;
                     }
-                } catch (\Exception $e) {
+                } catch (\Exception $e) { // @phpstan-ignore-line
                 }
 
                 if ($isMissing) {
--- a/src/DiContainerTrait.php
+++ b/src/DiContainerTrait.php
@@ -54,20 +54,36 @@ trait DiContainerTrait
             $setterName = 'set' . ucfirst($k);
             $setterExists = method_exists($this, $setterName) && $setterName !== 'setDefaults';
 
-            if ($setterExists || property_exists($this, $k)) {
-                if ($passively && ($setterExists ? $this->{$getterName}() : (isset($this->{$k}) && $this->{$k} !== null))) {
-                    continue;
+            if ($setterExists) { // when setter is declared, getter is expected to be declared too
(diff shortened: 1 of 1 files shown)


mvorisek:
@georgehristov wdyt?

DarkSide666:
Not sure about this.
@abbadon1334 comment https://github.com/atk4/core/issues/206#issuecomment-629648602 sounds quite legit.

georgehristov:
I am "for" it and was even proposing it in the chat some time ago.

mvorisek:
> 
> 
> Not sure about this.
> @abbadon1334 comment [#206 (comment)](https://github.com/atk4/core/issues/206#issuecomment-629648602) sounds quite legit.

we can implement it using 2pass approach, ie.
1. set DIRECTLY ALL values FIRST (if declared or detected to be magical)
2. then for each value (if setter exists):
    a) reset to old value
    b) set using invoking setXxx()

@abbadon1334 does this address your comment?

@DarkSide666 @georgehristov wdyt, any better idea?

mvorisek:
another option might be to use property anotations for getter/setter method names, but looks like quite a lot horsepower for the current state of this project now

src/DiContainerTrait.php
```diff
@@ -49,17 +49,48 @@ trait DiContainerTrait
...
+
+            if ($setterExists) { // when setter is declared, getter is expected to be declared too
+                $origValue = $this->{$getterName}();
+            } elseif (property_exists($this, $name)) {
+                $origValue = $this->{$name};
+            } else { // property may be magical
+                $isMissing = true;
+
+                try {
+                    $origValue = $this->{$name} ?? null;
```
mvorisek: this will still not work if null...

mvorisek:
closing in favor of https://wiki.php.net/rfc/property-hooks

PR #246, opened 2020-06-14
Allow hooks to be updated when active

### LINKED ISSUE #137: When hook for the same spot is added from hook itself, it is removed once the hook finish
https://github.com/atk4/core/blob/0d922e092f7ccd20efe29d197e84c5387d718565/src/HookTrait.php#L186

Is this desired and why?

mvorisek:
fix https://github.com/atk4/core/issues/137

### DIFF
--- a/src/HookTrait.php
+++ b/src/HookTrait.php
@@ -304,18 +304,48 @@ trait HookTrait
         if (isset($this->hooks[$spot])) {
             krsort($this->hooks[$spot]); // lower priority is called sooner
             $hooksBackup = $this->hooks[$spot];
+            $priorities = array_keys($hooksBackup);
+
             try {
-                while ($hooks = array_pop($this->hooks[$spot])) {
-                    foreach ($hooks as $index => [$hookFx, $hookArgs]) {
+                while (($priority = array_pop($priorities)) !== null) {
+                    $hooks2Backup = $this->hooks[$spot][$priority];
+                    $indexes = array_reverse(array_keys($hooks2Backup));
+
+                    while (($index = array_pop($indexes)) !== null) {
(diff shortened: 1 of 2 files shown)


src/HookTrait.php
```diff
@@ -107,6 +106,15 @@ private function _rebindHooksIfCloned(): void
         $this->_hookOrigThis = \WeakReference::create($this);
     }
 
+    private function _hookRequireInactive(string $spot): void
+    {
+        if ($this->_hookActiveSpots[$spot] ?? false) {
+            throw (new Exception('Hook spot must be inactive for requested operation, but it is already executing'))
```
mvorisek, line 112: TODO add exception test

src/HookTrait.php
```diff
@@ -305,13 +316,16 @@ public function hook(string $spot, array $args = [], HookBreaker &$brokenBy = nu
...
         if (isset($this->hooks[$spot])) {
-            krsort($this->hooks[$spot]); // lower priority is called sooner
-            $hooksBackup = $this->hooks[$spot];
+            ksort($this->hooks[$spot]); // lower priority is called sooner
+
             try {
-                while ($hooks = array_pop($this->hooks[$spot])) {
+                $this->_hookActiveSpots[$spot] = true;
+
+                foreach ($this->hooks[$spot] as $hooks) {
```
mvorisek, line 328:
```
$hookIndex = $app->onHook(App::HOOK_BEFORE_RENDER, static function () use ($fx, $app, &$hookIndex) {
    try {
        $fx();
    } finally {
        $app->removeHook(App::HOOK_BEFORE_RENDER, $hookIndex, true);
    }
});
```
might be wanted to be allowed

PR #247, opened 2020-06-18
Drop PHP 7.2 support

mvorisek:
now the time has ripen

to support better trait, nowdoc, ...

then apply as patch for all other major repos and update code with new features as well as CS rules


DarkSide666 (approved):
As agreed in meeting

mvorisek:
🚀 🚀 🚀 let's compete with SpaceX!

PR #248, opened 2020-06-20
Prevent item to be associated with two parents (prevent owner replace)

### LINKED ISSUE #237: Throw if adding object to a collection but it was already added to another
This is a feature to throw if object has owner already.

Currently when object is added, full name is computed. Therefore, adding object to an additional collection is currently possible, but in reality the is only one owner, one name, so if allowed, it is wrong.

mvorisek:
fix https://github.com/atk4/core/issues/237

tests for dsql/data/ui seems to be passing as well without any change needed (for these reasons not marked as breaking)

example in action:
image

image

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -7,6 +7,6 @@ coverage:
     project:
       default:
         target: auto
-        threshold: 0.1
+        threshold: 0.025
     patch: false
     changes: false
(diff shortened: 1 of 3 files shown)


mvorisek:
@georgehristov wdyt?

mvorisek:
@georgehristov @DarkSide666 LGTM?

DarkSide666:
Well... Need to think if there is at least one use-case when we actually would like ability to change objects owner.
I can't come up with any which can't be solved by cloning object and associate cloned object to another owner. That will be possible (clone object + add it to another owner)?

DarkSide666 (approved):

DarkSide666:
No use-case so far so i think it's fine.

PR #249, opened 2020-06-24
Require seed to be always an array or object

### LINKED ISSUE #226: Require seed to be always an array or object
will be breaking, but more for repos than user apps, because user's seed is typically an array

string seed - must be converted to array
null seed - must be converted to `[]`
false seed - usually used to not specify any - should be carefully checked and changed to `null`

array seed merged must be checked and decided which seeds should override what

mvorisek:
fixes: https://github.com/atk4/core/issues/226

merge after: https://github.com/atk4/data/pull/639
merge after: https://github.com/atk4/ui/pull/1315

why? currently, there is a lot of code that "prenormalize seed" and process it differently, example:
https://github.com/atk4/data/blob/8748a634b3ab4f7b02b5a009eaf2c28f917f3cb9/tests/ContainsOneTest.php#L54
notice, that it looks like a regular seed, but in reality the 1st argument/value is not an class name seed for that seed, but instead of for completely diffent class (and it has nothing to to with `caption`)

- `null` should be not supported as well (use `[]` instead), but `DIContainerTrait::setDefaults()` is not ready for this change now
- this PR means you are not allowed to use string seeds like `Cl::class`, but you have to use `[Cl::class]` instead
- other inputs types (like `int`, ...) may be working now, but only because of non-strict checking

In other words - as of now, non-object seed must be of `array` type with the first argument (value with numerical key) is class name, object seed remains object like before

Regex to find old usage:
```
(seed|factory)[^\n]*::class(?!.*\])
```
or more aggressive:
```
[=(][^\n\[]*(?<!exception)::class(?! \.)
```

data/ui PRs above may help you to understand what changes needs to be done

to debug it may help you to:
a) uncomment https://github.com/atk4/core/blob/9b9aad0dfa3105eb551bd5a479fa41cf9c17fa38/src/Factory.php#L49
b) replace if to `(true)` https://github.com/atk4/core/blob/9b9aad0dfa3105eb551bd5a479fa41cf9c17fa38/src/ExceptionRenderer/HTML.php#L158

### DIFF
--- a/src/DIContainerTrait.php
+++ b/src/DIContainerTrait.php
@@ -105,6 +105,8 @@ trait DIContainerTrait
                         ->addMoreInfo('seed_type', gettype($seed));
                 }
 
+                Factory::mergeSeeds($seed, []); // emit deprecated warning, remove in 2020-dec once string seed is no longer supported
+
                 $seed = [$seed];
             }
 
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -23,11 +23,45 @@ class Factory
         return self::$_instance;
     }
 
+    private function _printDeprecatedWarningWithTrace(string $msg): void // remove once not used within this class
+    {
+        static $traceRenderer = null;
+        if ($traceRenderer === null) {
+            $traceRenderer = new class(new Exception()) extends ExceptionRenderer\HTML {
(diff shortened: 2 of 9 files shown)


georgehristov (approved):
LGTM

PR #250, opened 2020-06-25
Release drafter to detect labels

romaninsh:
Consistency with ATK UI (https://github.com/atk4/ui/blob/develop/.github/release-drafter.yml)

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,5 +1,13 @@
-# See https://github.com/release-drafter/release-drafter#configuration
+categories:
+  - title: "Breaking Changes"
+    labels:
+      - "BC-break"
+  - title: "Major Features"
+    labels:
+      - "MAJOR"
+  - title: "Documentation enhancements"
+    labels:
+      - "Documentation :books:"
 template: |
   ## What’s Changed
-
   $CHANGES


.github/release-drafter.yml
```diff
@@ -1,5 +1,13 @@
-# See https://github.com/release-drafter/release-drafter#configuration
+categories:
+  - title: "Breaking Changes"
+    labels:
+      - "BC-break"
+  - title: "Major Features"
+    labels:
+      - "MAJOR"
+  - title: "Documentation enhancements"
```
mvorisek, line 8: does this help somehow?

It will be actually good if we can exclude `small-fix` labels (like for typos - documenting these PRs is unwanted) and maybe doc updates too

mvorisek:
This PR is not good, as if both `BC-break` and `MAJOR` labels are presented, the changes are duplicated under both sections.

Breaking changes should not be duplicated in any other section. Not sure if this is possible now.

We should exclude PRs for `small fix` labels, it should be possible - https://github.com/release-drafter/release-drafter#exclude-pull-requests

So maybe we can exlude all irelevant PRs instead and do not add major section at all.

mvorisek (approved):
for consistency LGTM

PR #251, opened 2020-06-26
Array seed requires valid (of omitted) class name

mvorisek:
intended to be merged into develop

- dsql: unaffected
- data: 2 lines of bad ui usage
- ui: see related PR

merge after: https://github.com/atk4/data/pull/640
merge after: https://github.com/atk4/ui/pull/1330

**This PR is part of very strict definition/impl. of seeds. Seed is defined now strictly as:**
- object which may be updated using DI (dependency injection), but the same instance is always returned
- an array:
  - which has arguments (values with numerical keys) and injection (KV pairs with non-numerical keys)
  - 1st argument must be string class name (or `null` for merging seeds) @TODO I think this is wrong too, as in the code we should be able to find classes in seeds easily, this should be probably implemented with some special class

**For BC, this PR does not restrict this:**
- throw if 1st argument is not present
- throw if 1st argument is null (see comment above, it maybe never will)
- seed is null (instead of `[]` - but DIContainerTrait::setDefaults() is not ready for this change now)

Regexes to find old usage:
`\[\s*new` - do not instance object for seed
-`\[[^,]+(?<!::class)(\]| ,)` - all usages of seeds where the seed class is omitted (very few cases across whole `ui` repo)

ui PR above may help you to understand what changes needs to be done

to debug it may help you to:
a) uncomment https://github.com/atk4/core/blob/9b9aad0dfa3105eb551bd5a479fa41cf9c17fa38/src/Factory.php#L49
b) replace if to `(true)` https://github.com/atk4/core/blob/9b9aad0dfa3105eb551bd5a479fa41cf9c17fa38/src/ExceptionRenderer/HTML.php#L158

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -142,6 +142,10 @@ class Factory
 
     protected function _factory($seed, $defaults = []): object
     {
+        if (is_object($defaults)) {
+            throw new Exception('Factory $defaults can not be an object');
+        }
+
         if ($defaults === null) { // should be deprecated soon
             $defaults = [];
         }
@@ -163,7 +167,7 @@ class Factory
 
         if (is_array($defaults)) {
             array_unshift($defaults, null); // insert argument 0
-        } elseif (!is_object($defaults)) {
+        } else {
             $defaults = [null, $defaults];
         }
         $seed = $this->_mergeSeeds($seed, $defaults);
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -172,12 +172,6 @@ class Factory
         }
         $seed = $this->_mergeSeeds($seed, $defaults);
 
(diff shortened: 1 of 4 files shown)


src/Factory.php
```diff
@@ -142,6 +174,10 @@ protected function _newObject(string $className, array $ctorArgs): object
 
     protected function _factory($seed, $defaults = []): object
     {
+        if (is_object($defaults)) {
```
mvorisek, line 177: see https://github.com/atk4/core/pull/252/files for changes in factory, not breaking...

mvorisek:
@georgehristov can you approve this?

georgehristov (approved):
LGTM

PR #252, opened 2020-06-27
Factory $defaults can not be an object

mvorisek:
expect two change in `core` tests, no changes needed for `data`, `dsql`, `ui`

marking as a bug, `addTo` does not accept object/class name by defintion (seed class name is defined with class before `::AddTo(`)


mvorisek:
closing in favor of https://github.com/atk4/core/pull/251

PR #253, opened 2020-06-29
Add testing with PHP 8.0

mvorisek:


romaninsh (approved):
few notes:
 - if something fails on 8.0 could be due to bugs in alpha release.
 - once 8.0 is released it would become 'latest' so we should introduce '7.4' instead

.github/workflows/unit-tests.yml
```diff
@@ -30,7 +30,7 @@ jobs:
 
       - name: Configure PHP
         run: |
-          if [ -z "$LOG_COVERAGE" ]; then rm /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini ; fi
+          if [ -z "$LOG_COVERAGE" ]; then rm /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini || true ; fi
```
romaninsh: could use to only return true for 8.0
```
.... || [ `php -v` -eq 8.0 ]
```

mvorisek: addressed, but otherwise prefer to test for GH action job var...

mvorisek:
> * if something fails on 8.0 could be due to bugs in alpha release.

I will merge once tests will be passing, now there is an issue with some dep that does not allow to use php 8.0 and requires phpunit 3.7 :D

> * once 8.0 is released it would become 'latest' so we should introduce '7.4' instead

yes and 8.0 can be probably kept

mvorisek:
waiting for https://github.com/sebastianbergmann/phpunit/issues/4325#issuecomment-651418974

should be resolved at August 7, 2020

PR #254, opened 2020-07-01
Try to relativize php paths in exception renderer

mvorisek:
Before:
image

After:
image


DarkSide666 (approved):
Looks good. Not sure why removed translator, but ... I guess it doesn't fully work anyway.

mvorisek:
I did not - moved here https://github.com/atk4/core/pull/254/files#diff-414b77429b6867930a869353a407494eR133

fallback message is not translated because of possible error

PR #255, opened 2020-07-02
Revert "Drop PHP 7.2 support"

mvorisek:
revert until one stable releases for php 7.2 are made, otherwise if the broken releases will be used by everyone still on php 7.2 and we do not want these complains...

after release, drop support forever


mvorisek:
@DarkSide666 thinks we do not have to fix this for php7.2 users...

PR #258, opened 2020-07-07
Fix PSR class/trait/method naming

mvorisek:
pure case refactor only, class/trait/method names are CI

renamed to pure PascalCase (camelCase for methods) with snake case reasoning (like you will never write `h_t_m_l`)

based on `(?<!::)(trait|interface|function|class) ((?!(Expression|Connection|Query)_).)*[A-Z][A-Z]`

the only breaking change is that these files will not load using autoloaded if not optimized and not using CI FS

updated names:
```
DIContainerTrait
atk4\core\ExceptionRenderer\HTML
atk4\core\ExceptionRenderer\JSON
Exception::getHTML()
Exception::getJSON()
```

for 100% BC you can require the renamed classes manually (but if some other updated code required them before use is enough)

### DIFF
--- a/docs/di.rst
+++ b/docs/di.rst
@@ -2,7 +2,7 @@
 Dependency Injection Container
 ==============================
 
-.. php:trait:: DIContainerTrait
+.. php:trait:: DiContainerTrait
 
 Agile Core implements basic support for Dependency Injection Container.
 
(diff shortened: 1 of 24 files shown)


mvorisek:
@georgehristov ok with you, can you approve?

georgehristov (approved):
For me it follows agreed naming convention so your effort on this is appreciated!

PR #262, opened 2020-07-10
Fix TestCase::callProtected for magic calls

mvorisek:


mvorisek:
@georgehristov ok with you?

DarkSide666 (approved):
LGTM

PR #263, opened 2020-07-18
Remove recursive from Factory::_mergeSeeds()

mvorisek:
100% BC, simpler and 30% faster (much faster when 3 or more seeds were passed, as they were reparsed/sorted again)

### DIFF
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -50,7 +50,8 @@ class Factory
         'trigger_error'($msg . (!class_exists(\PHPUnit\Framework\Test::class, false) ? "\n" . $trace : ''), E_USER_DEPRECATED);
     }
 
-    private function checkSeeFunc($seed): ?string {
+    private function checkSeeFunc($seed): ?string
+    {
         if (is_object($seed) || $seed === null) {
             return null;
         } elseif (!array_key_exists(0, $seed)) {
@@ -87,6 +88,7 @@ class Factory
             if (is_object($seed)) {
                 if ($obj !== null) {
                     continue; // legacy behaviour
+
                     throw new \Exception('Two or more objects specified as seed.');
                 }
 
@@ -128,6 +130,7 @@ class Factory
         ksort($arguments, SORT_NUMERIC);
         if ($obj === null) {
             $arguments = $arguments + $injection;
+
             return $arguments;
         }
 
@@ -155,7 +158,6 @@ class Factory
         }
 
         return $obj;
-
     }
 
     protected function _newObject(string $className, array $ctorArgs): object
--- a/src/Factory.php
+++ b/src/Factory.php
@@ -50,121 +50,112 @@ class Factory
         'trigger_error'($msg . (!class_exists(\PHPUnit\Framework\Test::class, false) ? "\n" . $trace : ''), E_USER_DEPRECATED);
     }
 
-    protected function _mergeSeeds($seed, $seed2, ...$more_seeds)
+    private function checkSeeFunc($seed): ?string {
+        if (is_object($seed) || $seed === null) {
+            return null;
+        } elseif (!array_key_exists(0, $seed)) {
+            return null; // 'not defined' allow this method to be used to merge seeds without class name
+        } elseif ($seed[0] === null) {
+            return null;
+        } elseif (!is_string($seed[0])) {
+            return 'invalid type (' . (is_object($seed[0]) ? get_class($seed[0]) . ' (class wrapped in an array?)' : gettype($seed[0])) . ')';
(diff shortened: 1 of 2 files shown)


tests/SeedTest.php
```diff
@@ -149,7 +149,8 @@ public function testMerge5()
         // works even if more arguments present
         $o = new ViewTestMock();
         $o->foo = ['red'];
-        $oo = $this->mergeSeeds(['foo' => ['xx']], ['foo' => ['green']], $o);
+        $oo = $this->mergeSeeds(['foo' => ['green']], $o);
```
mvorisek, line 152: test update is needed because resolved at merge time and injected only the final one...

src/Factory.php
```diff
@@ -50,121 +50,114 @@ public function tryRelativizePath(string $path): string
...
-        if (count($more_seeds) > 0) {
-            $seed2 = $this->_mergeSeeds($seed2, ...$more_seeds);
+        if (is_object($seed) || $seed === null) {
+            return null;
+        } elseif (!array_key_exists(0, $seed)) {
+            return null; // 'not defined' allow this method to be used to merge seeds without class name
+        } elseif ($seed[0] === null) {
+            return null;
+        } elseif (!is_string($seed[0])) {
+            return 'invalid type (' . (is_object($seed[0]) ? get_class($seed[0]) . ' (class wrapped in an array?)' : gettype($seed[0])) . ')';
```
DarkSide666, line 62: more proper message would be _**object** wrapped in an array_ not _**class** mapped in an array_.

src/Factory.php
```diff
@@ -50,121 +50,114 @@ public function tryRelativizePath(string $path): string
...
-            } elseif ($seed[0] === null) {
-                return null;
-            } elseif (!is_string($seed[0])) {
-                return 'invalid type (' . (is_object($seed[0]) ? get_class($seed[0]) . ' (class wrapped in an array?)' : gettype($seed[0])) . ')';
-            } elseif (class_exists($seed[0])) {
+        // do not emit warnings for core tests:
+        // - some tests already tests for exception
+        // - we may later want to use this function for "mergeDefaults" (like _factory() below does)
+        foreach (debug_backtrace(DEBUG_BACKTRACE_IGNORE_ARGS) as $cl) {
+            if (strpos($cl['class'] ?? '', 'atk4\core\tests\\') === 0) {
```
DarkSide666, line 71: Why this is needed here? Looks something wrong by design if such hack is needed here in code.

src/Factory.php
```diff
@@ -50,121 +50,114 @@ public function tryRelativizePath(string $path): string
...
-            if (is_array($seed2)) {
-                $arguments = array_filter($seed2, 'is_numeric', ARRAY_FILTER_USE_KEY); // with numeric keys
-                $injection = array_diff_key($seed2, $arguments); // with string keys
-                unset($seed2);
-                unset($arguments[0]); // the first argument specifies a class name
+            if ($this->checkSeeFunc($seed) !== null) {
+                // remove/do not accept other seed than object/array type after 2020-dec
+                // remove/do not accept seed with 1st argument other than valid class name (or null) after 2020-dec
+                $this->printDeprecatedWarningWithTrace(
+                    'Use of invalid/deprecated $seed' . $seedIndex . ' class name (' . $this->checkSeeFunc($seed) . '). Support will be removed shortly.'
```
DarkSide666, line 109: Probably not that important, but this calls checkSeedFunc() again.

tests/SeedTest.php
DarkSide666: But will this also work?
`$this->mergeSeeds(['foo' => ['xx']], ['foo' => ['green']], $o);` ?

DarkSide666 (changes requested):

tests/SeedTest.php
mvorisek: yes, but `foo` key is passed to setDefaults only once (previously this was not strict, as merged in pairs)

src/Factory.php
mvorisek: this is moved from inline function to private one only, code is there only for transition, this should be updated later

the good news is that we already check the seed quite strictly, so there should be no break in every release...

src/Factory.php
mvorisek: this is moved from inline function to private one only, code is there only for transition...

src/Factory.php
mvorisek: it is never called when code is ok ;-)

DarkSide666 (approved):
LGTM

PR #269, opened 2020-08-30
remove duplicate func_num_args == 2 check

PhilippGrashoff:
Remove this check as called function checks exactly the same:
```
//check in Factory.php

 final public static function factory($seed, $defaults = []): object
    {
        if (func_num_args() > 2) { // prevent bad usage
            throw new \Error('Too many method arguments');
        }

        return self::getInstance()->_factory($seed, $defaults);
    }
```

### DIFF
--- a/src/FactoryTrait.php
+++ b/src/FactoryTrait.php
@@ -33,10 +33,6 @@ trait FactoryTrait
      */
     public function factory($seed, $defaults = []): object
     {
-        if (func_num_args() > 2) { // prevent bad usage
-            throw new \Error('Too many method arguments, factory does no longer support prefix');
-        }
-
         return Factory::factory($seed, $defaults);
     }
 }


mvorisek:
@PhilippGrashoff but user can call the later function directly, right? :)

PhilippGrashoff:
@mvorisek don't really get what you mean. FactoryTrait calls Factory. The check in FactoryTrait would be removed because Factory does exactly the same check.

mvorisek:
> 
> 
> @mvorisek don't really get what you mean. FactoryTrait calls Factory. The check in FactoryTrait would be removed because Factory does exactly the same check.

`factory()` is public method and we need to/should check there func args as well

PhilippGrashoff:
Hi,

still do not get your argument:
FactoryTrait::factory() only does one thing: call Factory::factory(),
Factory::factory() checks number of args anyway.
So the check in FactoryTrait::factory() isn't needed, as it only calls Factory::factory(), which checks anyway.

mvorisek:
> FactoryTrait::factory() only does one thing: call Factory::factory(),

with your proposed change, calling `FactoryTrait::factory()` with 3 args will pass, which is unwanted

PhilippGrashoff:
Why should it pass? it calls Factory::factory which will throw error.

mvorisek:
because it checks passed args - https://github.com/atk4/core/pull/269/files#diff-0465835788d64e7f6e72e370ba840cfaL36

and we forward only two, so check later will pass

try yourself :)

PhilippGrashoff:
omg you are so right. Thanks! Thats what happens when not writing tests for code changes.

PR #272, opened 2020-09-01
Hook closure must be unbound or bound to the this when cloning

### LINKED ISSUE #212: Cloned hooks are not rebound
Feature request, not a bug. Mainly for highly cloned data `Model` class.

slightly related with: #201

Example:
```
class A {
public function __construct() {
  $this->fx = [$this, 'hook'];
}

private function hook() {
  var_dump(spl_object_hash($this));
}

public function run() {
  ($this->fx)();
}
}

$a = new A();
$a->run();

$b = clone $a;
$b->run();
```

Can be probably solved by implementing a trait with some cloneable closure support.

mvorisek:
fixes https://github.com/atk4/core/issues/212

merge once https://github.com/atk4/data/pull/711 is finished

(not a direct bug, but it leads to bugs when target object is cloned)

no BC break - because later https://github.com/atk4/core/pull/275 PR

this PR solves a cloning issue when `onHook` is called with bounded Closured to other object than `$this`, like in:
https://github.com/atk4/data/blob/9ef78eab556930994b388270cb01703674c08217/src/Reference/HasOne.php#L213

### DIFF
--- a/docs/hook.rst
+++ b/docs/hook.rst
@@ -30,7 +30,7 @@ The framework or application would typically execute hooks like this::
 
 You can register multiple call-backs to be executed for the requested `spot`::
 
-    $obj->onHook('spot', function($obj){ echo "Hook 'spot' is called!"; });
+    $obj->onHook('spot', function($obj) { echo "Hook 'spot' is called!"; });
 
 Adding callbacks
 ================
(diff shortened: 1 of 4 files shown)


mvorisek:
@georgehristov wdyt?

georgehristov (approved):
LGTM

PR #275, opened 2020-09-18
Make hook check less strict - check only for the same class name

mvorisek:
introduced in https://github.com/atk4/core/pull/272

for core/dsql/data repos, we can keep the check active as long as there is no strong usecase to disable it, this strick check can payoff with better cloning safety

only Dark can merge this one :)


DarkSide666 (approved):
`atk4/audit` tests pass with this change

PR #277, opened 2020-10-01
Move "app" and "owner" props to getter/setter

mvorisek:
there are at least 3 reasons why:
- have more control (type check, reassign, enforce presence in some situations like after init, ...)
- allow to use WeakReference (for owner to allow to release child object based on ref counting not requiring expensive/slow GC)
- make sure we do not get/set properties on object not "implementing" given traits

if AppScopeTrait trait is present, we may even want to assert if App is set in any init

### How to update your code?
1. replace `->owner->` with `->getOwner()->`
2. search for `->owner(?!\w)` and adjust to `getOwner()`, `setOwner()`, `issetOwner()` manually
3. do the same with app property

### DIFF
--- a/src/AppScopeTrait.php
+++ b/src/AppScopeTrait.php
@@ -100,6 +100,7 @@ trait AppScopeTrait
     public function getApp()
     {
         $this->assertNoDirectAppAssignment();
+        $this->assertInstanceOfApp($this->_app);
 
         return $this->_app;
     }
--- a/src/AppScopeTrait.php
+++ b/src/AppScopeTrait.php
@@ -115,7 +115,9 @@ trait AppScopeTrait
         $this->assertNoDirectAppAssignment();
         $this->assertInstanceOfApp($app);
         if ($this->issetApp() && $this->getApp() !== $app) {
-            throw new Exception('App can not be replaced');
+            if ($this->getApp()->catch_exceptions || $this->getApp()->always_run) { // allow to replace App created by AbstractView::initDefaultApp() - TODO fix
+                throw new Exception('App can not be replaced');
+            }
         }
 
         $this->_app = $app;
--- a/src/AppScopeTrait.php
+++ b/src/AppScopeTrait.php
@@ -114,8 +114,8 @@ trait AppScopeTrait
     {
         $this->assertNoDirectAppAssignment();
         $this->assertInstanceOfApp($app);
-        if ($this->issetApp()) {
-            throw new Exception('App already set');
+        if ($this->issetApp() && $this->getApp() !== $app) {
+            throw new Exception('App can not be replaced');
         }
 
         $this->_app = $app;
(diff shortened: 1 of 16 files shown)


src/AppScopeTrait.php
```diff
@@ -22,11 +24,11 @@ trait AppScopeTrait
...
 
     /**
-     * Always points to current Application.
+     * Always points to current application.
      *
-     * @var \atk4\ui\App
+     * @var App
      */
-    public $app;
+    private $app;
```
mvorisek: because of trait design, private prop will still be accessible from the class where used

we may want to rename to `_app`, but then any assigment will not throw - but it can be solved by checking `->app` inside init or in the getter/setter methods in this trait

src/DebugTrait.php
```diff
@@ -85,10 +85,10 @@ public function log($level, $message, array $context = [])
      */
     public function userMessage(string $message, array $context = [])
     {
-        if (isset($this->app) && $this->app instanceof \atk4\core\AppUserNotificationInterface) {
-            $this->app->userNotification($message, $context);
-        } elseif (isset($this->app) && $this->app instanceof \Psr\Log\LoggerInterface) {
-            $this->app->log('warning', 'Could not notify user about: ' . $message, $context);
+        if (isset($this->_appScopeTrait) && $this->getApp() instanceof \atk4\core\AppUserNotificationInterface) {
+            $this->getApp()->userNotification($message, $context);
+        } elseif (isset($this->_appScopeTrait) && $this->getApp() instanceof \Psr\Log\LoggerInterface) {
```
mvorisek, line 90: this was/is probably wrong and should be `$this->getApp()->logger`

src/AppScopeTrait.php
```diff
@@ -59,4 +66,43 @@ trait AppScopeTrait
...
+    }
+
+    public function issetApp(): bool
+    {
+        $this->assertNoDirectAssignment();
+
+        return $this->_app !== null;
+    }
+
+    public function getApp(): App
```
mvorisek: impossible to test if there is no App class - do we want this trait really in core? If yes, we can enforce object type and phpdoc App.

mvorisek:
@georgehristov PRs for all major repos are done, ok to merge?

romaninsh:
can this be backward compatible for 1 major release?

mvorisek:
> 
> 
> can this be backward compatible for 1 major release?

no, as it would require magic properties almost in every object

the good new is, that existing code can be migrated almost completely with "ctrl+h"

romaninsh:
@mvorisek I'd love if we avoid this. I mean you could add a dummy getOwner() that returns $this->owner for entire release for backward compatibility. Can we have that for 2.3 so that users could start updating code before migrating?

romaninsh (approved):
OK, but please add compatibility into 2.3 release branch.

PR #278, opened 2020-10-01
Drop PHP v7.3 support and upgrade CS Fixer to v3

mvorisek:
to support using `WeakReference` and other new features like `fn`, typed properties

### DIFF
--- a/.github/workflows/test-unit.yml
+++ b/.github/workflows/test-unit.yml
@@ -63,7 +63,7 @@ jobs:
 
       - name: Check Coding Style (only for CodingStyle)
         if: matrix.type == 'CodingStyle'
-        run: vendor/bin/php-cs-fixer fix --dry-run --using-cache=no --diff --diff-format=udiff --verbose --show-progress=dots
+        run: vendor/bin/php-cs-fixer fix --dry-run --using-cache=no --diff --verbose
 
       - name: Run Static Analysis (only for StaticAnalysis)
         if: matrix.type == 'StaticAnalysis'
@@ -79,7 +79,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php: ['7.3', '7.4', 'latest']
+        php: ['7.4', 'latest']
         type: ['Phpunit']
         include:
           - php: 'latest'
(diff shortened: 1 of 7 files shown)


mvorisek:
@DarkSide666 @georgehristov ok to drop php 7.3 for upcoming 2.4 release?

we can already make ˙8.0˙ the latest CI image and develop for php 7.4 and 8.0 versions only.

DarkSide666:
What advantage we get to deprecate 7.3 at this point?
I still use 7.3 quite a lot :)

mvorisek:
Mostly syntactic sugar like short closures, with php 8 static return type, ... I do not pursue this PR fow now.

Someday we might transpile code for lower php versions.

PhilippGrashoff:
Well, since 06.12.2020 there are only bugfixes, PHP 7.3:
https://www.php.net/supported-versions.php

At some point I think we decided to support PHP versions until they run out at php.net, so another year for PHP 7.3.

PhilippGrashoff:
Do I really remember correctly? We dropped PHP 7.2 before it ran out. However, dropping version support too soon will probably piss off users.

mvorisek:
I do not use PHP 7.3 anywhere, I support to drop PHP 7.3 immediatelly, let me know if we should merge :)

DarkSide666:
I think we should support 7.3 until it runs out completely in php.net.
Exception could be in case 7.3 somehow really limits us to continue work on quality toolkit, but if that's just few changes in syntax between 7.3 and 7.4 and no really impressive and useful change, then I see no point to drop support earlier than php.net does that.

mvorisek:
PHP 7.3 is already phased out from active development, let's remove the support for the upcoming atk4/* v3.0

PhilippGrashoff:
Agreeing on this, lets drop it for upcoming versions. This also means that we can use typed properties where possible.

mvorisek:
@PhilippGrashoff good point on typed props 👍

PhilippGrashoff:
looking very good at first glance, thx for the work!

georgehristov (approved):

PR #279, opened 2020-10-06
Add app/owner getter/setter to 2.3.x

mvorisek:
as requested in https://github.com/atk4/core/pull/277#issuecomment-704115840


romaninsh (approved):
Awesome!

PR #280, opened 2020-10-06
don't need this, it breaks composer on releases

romaninsh:
@DarkSide666 - is this needed? worked before it just fine. It's not helping with releases.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -29,7 +29,6 @@
       "homepage": "https://mvorisek.cz/"
     }
   ],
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -29,7 +29,6 @@
       "homepage": "https://mvorisek.cz/"
     }
   ],
-  "version": "dev-develop",
```
mvorisek, line 32: I belive this is highly needed to pass `composer install` in some situation when we are not at develop branch

PR #283, opened 2020-10-14
Update release builder

DarkSide666:
It does all release tasks except the updating "require". This can be done manually/later.

It can be triggered by pushing to "release/*" branch or to a branch suffixed with ".build".

It does then the changes a (force-)pushes the changes to branch with ".gen" suffix.

The release (on the generated branch) should be then checked and if ok, simply tagged with the release version.


.github/workflows/bundler.yml
```diff
@@ -28,17 +31,17 @@ jobs:
...
-          pr_title: "Releasing ${{ env.version }} into master"
-          pr_body: |
-            - [ ] Review changes (must include stable dependencies)
-            - [ ] Merge this PR into master (will delete ${{ github.ref }})
-            - [ ] Go to Releases and create TAG from master
-            Do not merge master into develop
-          pr_reviewer: "romaninsh"
-          pr_assignee: "romaninsh"
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+      #- name: pull-request
```
mvorisek: definitely do not comment out old code, simply remove the code

DarkSide666: yes i will cleanup/remove at the end before merging

.github/workflows/bundler.yml
```diff
@@ -22,10 +22,12 @@ jobs:
           # set proper version
           #composer config version --unset
           v=$(echo ${{ github.ref }} | cut -d / -f 4)
-          echo "::set-env name=version::$v"
+          #echo "::set-env name=version::$v"
```
mvorisek: we use simillar construction in unit testing: https://github.com/atk4/core/blob/develop/.github/workflows/unit-tests.yml#L43

composer.json
```diff
@@ -29,7 +29,7 @@
       "homepage": "https://mvorisek.cz/"
     }
   ],
-  "version": "dev-develop",
+  "version": "fix-version",
```
mvorisek: this is of course invalid to be commited into develop

DarkSide666: sorry, fixed

composer.json
```diff
@@ -30,6 +30,7 @@
     }
   ],
   "version": "dev-develop",
+  "version-release": "#release_version#",
```
mvorisek: I belive, we should simply replace "dev-develop" directly and drop `*-release` keys completely, it will simplify the configs and mistakes in all repos

DarkSide666: If replacing atk4 dev-develop is the only thing to replace, then maybe, but idea behind using `*-release` was that it can be used for any other possible changes which we need when releasing too.

mvorisek: I am quite sure yes, it is the only thing to replace. We should do that, to simplify that configs and prevent mistakes.

For development, there is `require-dev`, which should be enought :)

romaninsh: Agree with Michael and better use jq for editing

.github/workflows/bundler.yml
```diff
@@ -10,35 +10,26 @@ jobs:
...
+          ref: ${{ github.ref }}
       - run: echo ${{ github.ref }}
       - name: Update to stable dependencies
         run: |
           # replaces X keys with X-release keys
           jq '. as $in | reduce (keys_unsorted[] | select(endswith("-release")|not)) as $k ({}; . + {($k) : (($k + "-release") as $kr | $in | if has($kr) then .[$kr] else .[$k] end) } )' < composer.json > tmp && mv tmp composer.json
-          composer config version --unset
+
+          # set proper version
+          #composer config version --unset
```
mvorisek: If there is no reason I do not know, do rely on git tags.

mvorisek (approved):

PR #286, opened 2020-10-25
Suppress numerical keys from Exception renderer when canonical

mvorisek:


src/ExceptionRenderer/RendererAbstract.php
```diff
@@ -124,14 +124,15 @@ public static function toSafeString($val, $allowNl = false, int $maxDepth = 2):
         }
 
         $out = '[';
+        $supressKeys = array_keys($val) === range(0, count($val) - 1);
```
mvorisek, line 127: false for empty array, but ok here...

georgehristov (approved):
LGTM

PR #288, opened 2020-10-28
Fix renderer - do not access any magic property during rendering

mvorisek:
before:
image

after:
standard exception page (no core error)

### DIFF
--- a/src/ExceptionRenderer/RendererAbstract.php
+++ b/src/ExceptionRenderer/RendererAbstract.php
@@ -94,7 +94,10 @@ abstract class RendererAbstract
         }
 
         if ($parsed['object'] !== null) {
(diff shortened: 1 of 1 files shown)


src/ExceptionRenderer/RendererAbstract.php
```diff
@@ -94,7 +94,10 @@ protected function parseStackTraceCall(array $call): array
         }
 
         if ($parsed['object'] !== null) {
-            $parsed['object_formatted'] = $parsed['object']->name ?? get_class($parsed['object']);
+            $objProps = get_object_vars($parsed['object']);
```
mvorisek, line 97: get_object_vars does not touch magic in any sense

georgehristov (approved):
LGTM

PR #289, opened 2020-11-04
Deprecate FactoryTrait trait

mvorisek:
merge after https://github.com/atk4/data/pull/780 and https://github.com/atk4/ui/pull/1531

Factory is replaceable singleton.

implementing factory method using this to create another classes is nonsense


DarkSide666 (approved):
OK, makes sense.

PR #290, opened 2020-11-30
Add PHPStan to CI

mvorisek:
exclude rules for traits generated by:
```
foreach (scandir(__DIR__) as $fname) {
    if (!preg_match('~.*Trait\.php$~', $fname)) {
        continue;
    }

    $d = file_get_contents($fname);
    preg_match_all('~(?:private|protected|public)[^\n()]+(?:function[^\n()]+?(\w+)\(|\$(\w+))~', $d, $matchesAll, \PREG_SET_ORDER);
    $methods = [];
    $props = [];
    foreach ($matchesAll as $matches) {
        //print_r($matches);
        if ($matches[1] !== '') {
            $methods[] = $matches[1];
        } else {
            $props[] = $matches[2];
        }
    }

    echo '        # for ' . $fname . "\n";
    foreach ($methods as $n) {
        echo '        - \'~^Call to an undefined method .+::' . preg_quote($n, '~') . '\(\)\.$~\'' . "\n";
    }
    foreach ($props as $n) {
        echo '        - \'~^Access to an undefined property .+::\$' . preg_quote($n, '~') . '\.$~\'' . "\n";
    }
}
```

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -32,7 +32,7 @@ jobs:
 
       - name: Configure PHP
         run: |
-          if [ -z "$LOG_COVERAGE" ]; then rm /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini ; fi
+          if [ -n "$LOG_COVERAGE" ]; then echo "xdebug.mode=coverage" >> /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini; else rm /usr/local/etc/php/conf.d/docker-php-ext-xdebug.ini; fi
           php --version
 
       # trick composer that this is a "atk4/core:develop" dependency to install atk4/data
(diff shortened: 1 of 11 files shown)


mvorisek:
for
```
 ------ ------------------------------------------------------------------------------ 
  Line   src/CollectionTrait.php (in context of class atk4\core\tests\CollectionMock)  
 ------ ------------------------------------------------------------------------------ 
  51     Call to an undefined method                                                   
         $this(atk4\core\tests\CollectionMock)::getApp().   
```

and simillar errors I think we need interfaces to check traits (or customize phpstan)

@georgehristov wdyt?

georgehristov:
As we are on php 7.3 already we can use traits within traits already. php 7.2 had issues if these duplicate.
So in this case in CollectionTrait we can add `use AppScopeTrait`

mvorisek:
> 
> 
> As we are on php 7.3 already we can use traits within traits already. php 7.2 had issues if these duplicate.
> So in this case in CollectionTrait we can add `use AppScopeTrait`

`in this case` :)

we of course needs to be able to support
```
if (something understandable by phpstan, language provides one way - interfaces, like $x instanceof AppScopeTraitInterface) {
    $x->getApp(); // phpstan must be sure $x has AppScopeTrait
}
```

georgehristov:
Interfaces is the most correct way, agree

DarkSide666 (approved):
LGTM

PR #291, opened 2020-12-03
Update NS to CamelCase

mvorisek:
see https://github.com/atk4/data/pull/795


georgehristov (approved):
Ideally we would rename also core to Core and atk4 to something more catchy 😄

PR #293, opened 2020-12-03
Fix for CamelCase NS update

mvorisek:


romaninsh (approved):
TEMPORARILY APPROVING, but CamelCase policy adoption for the namespaces must be properly put to vote and unless welcomed, would be reverted throughout all repositories.

PR #304, opened 2020-12-28
Remove Check for Atk4\Ui\App in AppScopeTrait

abbadon1334:
Hello guys,
I came back and I test on one of my old projects and nothing worked.
I made this PR to solve the issue.

Ok for checking if is already set and direct set, is good, but not the check of the type.

I think we need to clarify that was passed a wrong concept about Atk4\Core.
Atk4/Core was created as an independent package and must not be hard coupled to others of Atk4.

The property app in Atk4\Core\AppScopeTrait is an "abstract app centralizer of service" used as a container, IS NOT a reference to \Atk4\Ui\App, with that type check the use of Atk out of Atk packages became impossible.

I fix some inspection errors.

### DIFF
--- a/src/AppScopeTrait.php
+++ b/src/AppScopeTrait.php
@@ -86,7 +86,6 @@ trait AppScopeTrait
     public function getApp()
     {
         $this->assertNoDirectAppAssignment();
-        $this->assertInstanceOfApp($this->_app);
 
         return $this->_app;
     }
@@ -99,7 +98,6 @@ trait AppScopeTrait
     public function setApp(object $app)
     {
         $this->assertNoDirectAppAssignment();
-        $this->assertInstanceOfApp($app);
         if ($this->issetApp() && $this->getApp() !== $app && $this->getApp() instanceof \Atk4\Ui\App) {
             if ($this->getApp()->catch_exceptions || $this->getApp()->always_run) { // allow to replace App created by AbstractView::initDefaultApp() - TODO fix
                 throw new Exception('App can not be replaced');
--- a/src/AppScopeTrait.php
+++ b/src/AppScopeTrait.php
@@ -63,18 +63,6 @@ trait AppScopeTrait
      */
     public $unique_hashes = [];
 
-    private function assertInstanceOfApp(object $app): void
-    {
-        // called from phpunit, allow to use/test this trait without \Atk4\Ui\App class
-        if (class_exists(\PHPUnit\Framework\TestCase::class, false)) {
-            return;
-        }
-
-        if (!$app instanceof \Atk4\Ui\App) {
-            throw new Exception('App must be instance of \Atk4\Ui\App');
-        }
-    }
-
     /**
      * To be removed in Jan 2021.
      */
@@ -112,7 +100,7 @@ trait AppScopeTrait
     {
         $this->assertNoDirectAppAssignment();
         $this->assertInstanceOfApp($app);
-        if ($this->issetApp() && $this->getApp() !== $app) {
+        if ($this->issetApp() && $this->getApp() !== $app && $this->getApp() instanceof \Atk4\Ui\App) {
             if ($this->getApp()->catch_exceptions || $this->getApp()->always_run) { // allow to replace App created by AbstractView::initDefaultApp() - TODO fix
                 throw new Exception('App can not be replaced');
             }
--- a/src/AppScopeTrait.php
+++ b/src/AppScopeTrait.php
@@ -97,7 +97,6 @@ trait AppScopeTrait
      */
     public function setApp(object $app)
     {
-        $this->assertNoDirectAppAssignment();
         if ($this->issetApp() && $this->getApp() !== $app && $this->getApp() instanceof \Atk4\Ui\App) {
             if ($this->getApp()->catch_exceptions || $this->getApp()->always_run) { // allow to replace App created by AbstractView::initDefaultApp() - TODO fix
                 throw new Exception('App can not be replaced');
--- a/src/AppScopeTrait.php
+++ b/src/AppScopeTrait.php
@@ -97,7 +97,7 @@ trait AppScopeTrait
      */
     public function setApp(object $app)
     {
-        if ($this->issetApp() && $this->getApp() !== $app && $this->getApp() instanceof \Atk4\Ui\App) {
+        if ($this->issetApp() && $this->getApp() !== $app && is_a($this->getApp(),"\\Atk4\\Ui\\App", true)) {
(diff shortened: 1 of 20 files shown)


src/AtkPhpunit/TestCase.php
```diff
@@ -17,20 +17,20 @@ class TestCase extends \PHPUnit\Framework\TestCase
      *
      * @return mixed
      */
-    public function &callProtected(object $obj, string $name, ...$args)
+    public function callProtected(object $obj, string $name, ...$args)
```
mvorisek, line 20: do you understand what you are doing?

mvorisek, line 20: this is for reference there, anyway, we do not need to assing for return

src/ContainerTrait.php
```diff
@@ -58,8 +58,7 @@ public function add($obj, $args = []): object
     {
         if (is_array($args)) {
             $args1 = $args;
-            unset($args1['desired_name']);
-            unset($args1[0]);
+            unset($args1['desired_name'], $args1[0]);
```
mvorisek, line 61: orig unset is fine, better readable

src/ConfigTrait.php
```diff
@@ -65,7 +65,7 @@ public function readConfig($files = ['config.php'], string $format = 'php')
 
                     break;
                 case 'json':
-                    $tempConfig = json_decode(file_get_contents($file), true);
+                    $tempConfig = json_decode(file_get_contents($file), true, 512, JSON_THROW_ON_ERROR);
```
mvorisek, line 68: 👍

src/ContainerTrait.php
```diff
@@ -179,8 +177,7 @@ public function removeElement($short_name)
      */
     protected function _shorten(string $desired): string
     {
-        if (isset($this->_appScopeTrait)
-            && isset($this->getApp()->max_name_length)
+        if (isset($this->_appScopeTrait, $this->getApp()->max_name_length)
```
mvorisek, line 180: no merged isset, so much harder to read!

src/DebugTrait.php
```diff
@@ -85,10 +85,15 @@ public function log($level, $message, array $context = [])
      */
     public function userMessage(string $message, array $context = [])
     {
-        if (isset($this->_appScopeTrait) && $this->issetApp() && $this->getApp() instanceof \Atk4\Core\AppUserNotificationInterface) {
-            $this->getApp()->userNotification($message, $context);
-        } elseif (isset($this->_appScopeTrait) && $this->issetApp() && $this->getApp() instanceof \Psr\Log\LoggerInterface) {
-            $this->getApp()->log('warning', 'Could not notify user about: ' . $message, $context);
+        if (isset($this->_appScopeTrait) && $this->issetApp()) {
+            switch (true) {
```
mvorisek, line 89: switch (no)

src/Factory.php
```diff
@@ -47,7 +47,7 @@ public function tryRelativizePath(string $path): string
             return $traceRenderer->tryRelativizePath($matches[0]);
         }, $trace);
         // echo (new Exception($msg))->getHtml();
-        'trigger_error'($msg . (!class_exists(\PHPUnit\Framework\Test::class, false) ? "\n" . $trace : ''), E_USER_DEPRECATED);
+        trigger_error($msg . (!class_exists(\PHPUnit\Framework\Test::class, false) ? "\n" . $trace : ''), E_USER_DEPRECATED);
```
mvorisek, line 50: this is intented for some reason

src/Factory.php
```diff
@@ -89,7 +89,7 @@ protected function _mergeSeeds(...$seeds)
                 if ($obj !== null) {
                     continue; // legacy behaviour
 
-                    throw new \Exception('Two or more objects specified as seed.');
+                    //throw new \Exception('Two or more objects specified as seed.');
```
mvorisek, line 92: no :)

src/Factory.php
```diff
@@ -203,7 +203,6 @@ protected function _factory($seed, $defaults = []): object
 
             $seed = $this->_mergeSeeds($seed, $defaults);
         }
-        unset($defaults);
```
mvorisek, line 206: no :)

src/AppScopeTrait.php
```diff
@@ -63,18 +63,6 @@ trait AppScopeTrait
      */
     public $unique_hashes = [];
 
-    private function assertInstanceOfApp(object $app): void
```
mvorisek, line 66: this is there for a reason, you want to assing app that is not `\Atk4\Ui\App` based?

src/Factory.php
abbadon1334: but why add code needed for test in production code?
i see some here and there and is wrong

src/Factory.php
abbadon1334: that exception will never throw.

src/Factory.php
abbadon1334: you will not use that $defaults further and is not by reference.

src/Factory.php
mvorisek: to allow smother migration between versions

lifecycle:
- old feature
- deprecate
- drop completely

src/Factory.php
mvorisek: so we keep it until we remove `                    continue; // legacy behaviour`

keep to still support refactoring!

src/Factory.php
mvorisek: yes, but for explicitly stating that we do not want any use in the future :)

mvorisek:
basically start from scratch and propose only the change you want, for CS changes, create another PR, but it will be likely rejected

src/AppScopeTrait.php
abbadon1334: yes of course, why not? I use the AppScope without using Atk4/Ui, I use it to centralize service in a master class as an App and use even Container + Trackable + Initializer + Hook without any problem to apply a composition pattern and in any element have a reference to a master class which is well organized and not static

mvorisek: this is problematic because of hinted type for static analysis

I think we need to move this trait to `atk4/ui` or make another trait based on this one there

but what is the reason for it then in core?

abbadon1334: I know traits gives that type of problem, because Traits are wrong, private methods in Traits are much wrong.
Like i said in private, i like the work you are doing and the direction of "be much strict" but not with traits, be much strict using traits is wrong.
If you want to go further in this direction why not start thinking about using interfaces?

PR #305, opened 2020-12-28
Change visibility of method assertInstanceOfApp

abbadon1334:
allow override of method.

### DIFF
--- a/src/AppScopeTrait.php
+++ b/src/AppScopeTrait.php
@@ -63,7 +63,7 @@ trait AppScopeTrait
      */
     public $unique_hashes = [];
 
-    private function assertInstanceOfApp(object $app): void
+    protected function assertInstanceOfApp(object $app): void
     {
         // called from phpunit, allow to use/test this trait without \Atk4\Ui\App class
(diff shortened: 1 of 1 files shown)


PhilippGrashoff:
Like it, had some trouble with other private methods lately.

src/AppScopeTrait.php
```diff
@@ -63,7 +63,7 @@ trait AppScopeTrait
      */
     public $unique_hashes = [];
 
-    private function assertInstanceOfApp(object $app): void
+    protected function assertInstanceOfApp(object $app): void
```
mvorisek, line 66: https://3v4l.org/2qPLn :D

please try itself before

mvorisek (changes requested):

src/AppScopeTrait.php
abbadon1334: i need to override it in class not use it as it is, did you know what are you doing?

If i have to rewrite even traits, user can use another framework, no? I like private but not in traits, too much trouble. machine cycle have a cost.

mvorisek (approved):
well, makes sense that it is not possible to overwrite it easily for framework classes otherwise

PR #310, opened 2021-04-05
Add proper interfaces for traits

mvorisek:
and remove `TraitUtil` introduced in https://github.com/atk4/core/pull/227

https://github.com/php/php-src/pull/5708 would be helpful ;-)


mvorisek:
Reopen once protected interface methods are supported or there is a need to specify these interfaces as a type.

PR #320, opened 2021-05-12
Use WeakReference for owner

mvorisek:
improve PHP GC performance

## BC break:

if child is passed/assigned but the owner is no longer referenced, the owner is (can be) released implicitly by PHP GC


mvorisek:
@georgehristov @DarkSide666 data/ui tests are failing, what do you think about this feature?

mvorisek:
too complicated to use

PR #322, opened 2021-05-14
Remove trailing whitespaces from text files

mvorisek:
bash command used:

```
find . -not \( -name .git -prune \) -type f -exec grep -Iq . {} \; -print0 | xargs -0 sed -i -E "s/[[:space:]]*$//"
```


mvorisek:
closing in favor of https://github.com/atk4/data/pull/897

PR #323, opened 2021-06-21
Rename AtkPhpunit NS and zero memory leaks testing

mvorisek:
related with https://github.com/sebastianbergmann/phpunit/issues/4705


abbadon1334 (approved):
Better this way than mine with cycle. Good job as always @mvorisek

PR #324, opened 2021-08-25
Test with PHP 8.1

mvorisek:
merge once PHP 8.1 is stable

### DIFF
--- a/.github/workflows/build-release.yml
+++ b/.github/workflows/build-release.yml
@@ -12,7 +12,7 @@ jobs:
     name: Build Release
     runs-on: ubuntu-latest
     container:
-      image: atk4/image:latest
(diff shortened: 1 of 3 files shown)


.github/workflows/build-release.yml
```diff
@@ -12,7 +12,7 @@ jobs:
     name: Build Release
     runs-on: ubuntu-latest
     container:
-      image: atk4/image:latest
+      image: ghcr.io/mvorisek/image-php:latest
```
DarkSide666, line 15: can't we have this in atk4 and if needed do changes there?

DarkSide666 (approved):

.github/workflows/build-release.yml
mvorisek: image

PR #329, opened 2021-10-03
Warn if property does not exist and add it to DiContainerTrait

mvorisek:
This highly improves code quality as something like:

```
$m->noLongerSupportedProperty = 'x';
```
or
```
$m->eckbox= 'x'; // notice the typo
```

will warn if the new trait is added. To add it to the most of the code, we add it to `DiContainerTrait` by default.

The deprecation warnings can be supressed by a custom error handler or by overriding the `warnIfPropertyDoesNotExist` method.

### DIFF
--- a/src/DiContainerTrait.php
+++ b/src/DiContainerTrait.php
@@ -32,6 +32,8 @@ namespace Atk4\Core;
  */
 trait DiContainerTrait
 {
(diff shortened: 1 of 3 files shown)


abbadon1334 (approved):
LGTM

PR #333, opened 2021-11-05
Display short name as a fallback in exception render

mvorisek:
display name of Field for example

### DIFF
--- a/src/ExceptionRenderer/RendererAbstract.php
+++ b/src/ExceptionRenderer/RendererAbstract.php
@@ -114,7 +114,7 @@ abstract class RendererAbstract
         if ($val instanceof \Closure) {
             return 'closure';
         } elseif (is_object($val)) {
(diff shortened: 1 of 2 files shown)


src/ExceptionRenderer/RendererAbstract.php
```diff
@@ -114,7 +114,7 @@ public static function toSafeString($val, bool $allowNl = false, int $maxDepth =
         if ($val instanceof \Closure) {
             return 'closure';
         } elseif (is_object($val)) {
-            return get_class($val) . (\Atk4\Core\TraitUtil::hasTrackableTrait($val) ? ' (' . get_object_vars($val)['name'] . ')' : '');
+            return get_class($val) . (\Atk4\Core\TraitUtil::hasTrackableTrait($val) ? ' (' . (get_object_vars($val)['name'] ?? get_object_vars($val)['short_name']) . ')' : '');
```
DarkSide666, line 117: Probably would be good to ```use \Atk4\Core\TraitUtil;``` and here just ```TraitUtil```

DarkSide666 (approved):

PR #334, opened 2021-11-08
Fix dynamic method exception when method is defined but not visible

mvorisek:
match PHP behaviour, see https://3v4l.org/EKcDHc

### DIFF
--- a/docs/container.rst
+++ b/docs/container.rst
@@ -172,9 +172,9 @@ Container Trait
             $parent->removeElement('child1');
         }
 
-        $parent->each(function($child) {
+        foreach ($parent as $child) {
             $child->doSomething();
-        });
+        }
 
     Child object names will be derived from the parent name.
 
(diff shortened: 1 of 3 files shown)


docs/index.rst
```diff
@@ -194,8 +194,8 @@ class to define a business object, such as - ShoppingBag::
         protected function init(): void {
             parent::init();
 
-            $this->hasOne('user_id', new User());
-            $this->hasMany('Items', new Item())
+            $this->addHasOne('user_id', new User());
```
DarkSide666: these changes shouldn't be in this PR

DarkSide666 (changes requested):
addHasMany / HasOne changes shouldn't be in this PR (see docs/factory.rst)

docs/index.rst
mvorisek: I updated the docs here to match the latest changes, there should not be any of these in atk4/core actually :D

DarkSide666 (approved):

PR #341, opened 2022-02-08
Remove SessionTrait trait

mvorisek:
Session is solely atk4/ui feature. Session trait with read/update methods should be kept, but the handling should be moved to a separate `App\Session` class.


DarkSide666:
Why remove from core? If session trait is defined in core then we can use it also in projects which don't use atk4/ui.
What advantage is to remove it from core?

mvorisek:
> we can use it also in projects which don't use atk4/ui

what is the real usecase?

see atk4/ui PR, it does not even make sense to have traited, eg. uncentralized session management

PR #362, opened 2022-08-26
Enable phpstan bleading edge and strict rules

mvorisek:


src/ContainerTrait.php
```diff
@@ -106,7 +106,7 @@ protected function _addContainer(object $element, array $args): void
         $element->setOwner($this);
         $element->shortName = $name;
         if (TraitUtil::hasTrackableTrait($this) && TraitUtil::hasNameTrait($this) && TraitUtil::hasNameTrait($element)) {
-            $element->name = $this->_shorten((string) $this->name, $element->shortName, $element->name);
+            $element->name = $this->_shorten($this->name ?: '', $element->shortName, $element->name); // @phpstan-ignore-line
```
mvorisek, line 111: `(?<!\w)name'?[^,]+false` is needed for atk4/ui

PR #392, opened 2023-09-23
Allow psr/log v2 and v3 dep

gowrav-vishwakarma:
BC break: `void` is newly returned as per the PSR logging interface

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
     "require": {
         "php": ">=7.4 <8.4",
         "benmorel/weakmap-polyfill": "^0.4.0",
-        "psr/log": "^1.0",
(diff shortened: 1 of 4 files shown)


composer.json
```diff
@@ -32,7 +32,7 @@
     "require": {
         "php": ">=7.4 <8.4",
         "benmorel/weakmap-polyfill": "^0.4.0",
-        "psr/log": "^1.0",
+        "psr/log": "^3.0",
```
mvorisek:
```suggestion
        "psr/log": "^1.0 || ^2.0 || ^3.0",
```

we cannot limit the minimal version

gowrav-vishwakarma: Agreed.

PR #421, opened 2025-02-07
Fix console exception renderer

### LINKED ISSUE #420: ExceptionRenderer/Console switch to yellow font
https://github.com/atk4/core/blob/develop/src/ExceptionRenderer/Console.php

I guess something is wrong with this because it switch console font color to yellow after it finish dumping out text.
I expect it to switch back to previous text color as it was before dumping out this.

Is that possible?

DarkSide666:
fix #420

* fix missing reset codes
* add all colors as constants
* use unified text styling method
* replace PHP_EOL with \n because it works nicely in Windows console too
* add additional tests

### DIFF
--- a/src/ExceptionRenderer/Console.php
+++ b/src/ExceptionRenderer/Console.php
@@ -130,25 +130,27 @@ class Console extends RendererAbstract
                 $inAtk = false;
             }
 
+            $functionColor = $escapeFrame ? self::COLOR_RED : self::COLOR_YELLOW;
+
             $tokens = [
                 '{FILE}' => str_pad(mb_substr($call['file_rel'], -40), 40, ' ', \STR_PAD_LEFT),
                 '{LINE}' => str_pad($call['line'], 4, ' ', \STR_PAD_LEFT),
                 '{OBJECT}' => $call['object'] !== null ? ' - ' . $this->text($call['object_formatted'], [self::COLOR_GREEN]) : '',
                 '{CLASS}' => $call['class'] !== null ? $this->text($call['class_formatted'] . '::', [self::COLOR_GREEN]) : '',
-                '{FUNCTION}' => $call['function'] !== null ? $this->text($call['function'], [$escapeFrame ? self::COLOR_RED : self::COLOR_YELLOW]) : '',
+                '{FUNCTION}' => $call['function'] !== null ? $this->text($call['function'], [$functionColor]) : '',
             ];
 
             if ($index === 'self') {
                 $tokens['{FUNCTION_ARGS}'] = '';
             } elseif (count($call['args']) === 0) {
-                $tokens['{FUNCTION_ARGS}'] = '()';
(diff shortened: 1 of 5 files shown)


src/ExceptionRenderer/Console.php
```diff
@@ -134,12 +140,17 @@ protected function processPreviousException(): void
...
+    }
 
-        $this->output .= (string) (new static($this->exception->getPrevious(), $this->adapter, $this->exception));
-        $this->output .= <<<'EOF'
-            \e[1;31m--
-            \e[0m
-            EOF;
+    /**
+     * Style text with ASCII colors.
+     */
```
mvorisek:
```suggestion
    /**
     * @param non-empty-list<self::FORMAT_*|self::COLOR_*> $formats
     */
```

mvorisek: please avoid too much comments as this is even completely wrong, no human nor AI would be able to understand what it really does :)

tests/ExceptionTest.php
```diff
@@ -46,6 +46,28 @@ public function testBasic(): void
...
+        self::assertStringNotContainsString('\e[', $ret);
+        self::assertMatchesRegularExpression('~\e\[0m(?!.*\e\[[0-9;]*m)~s', $ret); // reset colors at the end
+
+        $m->addMoreInfo('a1', 111);
+        $ret = $m->getColorfulText();
+        self::assertStringContainsString("\e[", $ret);
+        self::assertStringNotContainsString('\e[', $ret);
+        self::assertMatchesRegularExpression('~\e\[0m(?!.*\e\[[0-9;]*m)~s', $ret); // reset colors at the end
+
+        $m->addSolution('Simple solution');
```
mvorisek: add all at once, text would be then deduplicated by itself ;-)

DarkSide666: Not exactly. First I test simple exception, then test with added additional info and then with added solution info.

src/ExceptionRenderer/Console.php
```diff
@@ -134,12 +142,17 @@ protected function processPreviousException(): void
...
 
-        $this->output .= (string) (new static($this->exception->getPrevious(), $this->adapter, $this->exception));
-        $this->output .= <<<'EOF'
-            \e[1;31m--
-            \e[0m
-            EOF;
+    /**
+     * @param non-empty-list<self::FORMAT_*|self::COLOR_*|self::BG_COLOR_*> $formats
+     */
+    private static function text(string $text, array $formats = []): string
```
mvorisek, line 154:
```suggestion
    private static function text(string $text, array $formats): string
```

tests/ExceptionTest.php
mvorisek: Ahh, I know what you mean.

We can newly assert format tokens are in pairs :) - see what I meant the test is too specific..

DarkSide666: Sorry I don't get what you want to do differently here :)
Maybe I'm to sleepy :)

PR #422, opened 2025-02-09
Improve exception renderers

mvorisek:
improve #421

### DIFF
--- a/src/ExceptionRenderer/Console.php
+++ b/src/ExceptionRenderer/Console.php
@@ -20,13 +20,15 @@ class Console extends RendererAbstract
     private const COLOR_BRIGHT_GREEN = "\e[92m";
 
     /**
-     * @param non-empty-list<self::FORMAT_*|self::COLOR_*|self::BACKGROUND_COLOR_*> $formats
+     * @param list<self::FORMAT_*|self::COLOR_*|self::BACKGROUND_COLOR_*> $formats
      */
-    private function text(string $text, array $formats): string
+    private function text(string $text, array $formats = []): string
     {
-        assert(!str_contains($text, "\e["));
+        $text = str_replace("\e", '', $text);
 
-        return implode('', $formats) . $text . self::RESET;
+        return $formats === []
+            ? $text
+            : implode('', $formats) . $text . self::RESET;
     }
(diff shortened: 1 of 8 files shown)


src/ExceptionRenderer/Console.php
```diff
@@ -20,21 +20,31 @@ class Console extends RendererAbstract
...
 
-        return implode('', $formats) . $text . self::RESET;
+        return $formats === []
+            ? $text
+            : implode('', $formats) . $text . self::RESET;
+    }
+
+    private function optimizeText(string $value): string
+    {
+        $res = preg_replace("~\e\\[\\d{1,2}m\e\\[0m~", '', $value);
```
DarkSide666, line 36: would be nice to add small comment here about what these regexes does
first removes empty sequences, but second ...

src/ExceptionRenderer/Json.php
```diff
@@ -96,7 +92,7 @@ protected function processStackTraceInternal(): void
                 }, $call['args']);
             }
 
-            $this->json['stack'][] = $call;
+            $this->json['trace'][] = $call;
```
DarkSide666, line 95: nice catch!

src/ExceptionRenderer/RendererAbstract.php
```diff
@@ -268,14 +268,12 @@ protected function makeRelativePath(string $path): string
 
     protected function tryRelativizePathsInString(string $str): string
     {
-        $str = preg_replace_callback('~(?<!\w)(?:[/\\\]|[a-z]:)\w?+[^:"\',;]*?\.php(?!\w)~i', function ($matches) {
+        return preg_replace_callback('~(?<!\w)(?:[/\\\]|[a-z]:)\w?+[^:"\',;]*?\.php(?!\w)~i', function ($matches) {
```
DarkSide666, line 271: would be nice to add comment about what this regex does

mvorisek: AFAIK, this is even untested (exact behaviour is unasserted), thus help welcomed. It is a little problematic as the repo can be part of another repo and the tests should still pass.

